Issue #37, opened 2015-11-18
Crawling Processes Page information getting into failure url when you click on any of the buttons there (Typo3 7.6.0)


julianstefanov:
When you get to the Backend Info page, then select the Site Crawler information page and then select the "Crawling Processes" from the Page information select box you got 3 buttons:
"Refresh", "Enable crawling" and "Show finished and terminated processes".
If you click on any of them, the right typo3 backend iframe window loading the index of the typo3 backend.
Look at the screenshots:
1 screen: the exact page where the buttons are
2 screen: the result when you click on any of them

![1](https://cloud.githubusercontent.com/assets/12820618/11236965/3572cbbc-8ddd-11e5-908d-71d4bece5a1b.png)

![2](https://cloud.githubusercontent.com/assets/12820618/11236966/39cc76cc-8ddd-11e5-8180-b810a37ef783.png)

tomasnorre:
Thanks for letting us know, I'll take a look.

tomasnorre:
I can confirm this issue.. I'll dig more into it. 
Thanks for reporting.

PR #38, opened 2015-11-27
[BUG] Fix Crawling buttons for TYPO3 7 LTS

tomasnorre:
Fixes issues #37


tomasnorre:
This is another suggestion to this i saw after this.. 

I think it might be a cleaner approach.. 
https://github.com/AOEpeople/crawler/commit/9b1dfb4f82412acb386256b73ebbbac459af9134#commitcomment-14381573

tomasnorre:
Another will be provided.

Issue #40, opened 2015-12-02
Not working with TYPO3 7.6 / usage of invalid functions


SventB:
The crawler ext. uses function \TYPO3\CMS\Core\Utility\GeneralUtility::loadTCA but this function is not available in TYPO3 version 7.6

tomasnorre:
Thanks. We will update

PR #41, opened 2015-12-08
[TASK] Remove usage of invalid functions

tomasnorre:
Fixes issue #40

According to this page: https://docs.typo3.org/typo3cms/extensions/core/latest/Changelog/7.0/Breaking-61785-LoadTcaFunctionRemoved.html this function call can be removed completely.


srotsch:
Closes #40

Issue #45, opened 2016-01-20
Crawling Processes Page Information - Refresh button does not work as expected in TYPO3 6.2


vdespa:
This is related to the change made in relation to this issue #37 (52c7b2d99d6d9e29b32506c19c3652cdd1ebb961).

The commit introduces a regression. 

Steps to reproduce in crawler version 5.0.1, TYPO3 6.2 or 7.6.2:

1. Add pages to the crawler queue (5-10 pages)
2. Manually start crawling processes (Add process)
3. Click reload (maybe multiple times).
4. When the processes are finished and exit, you will start seeing the error (because instead of refreshing you keep adding processes):

![screen shot 2016-01-20 at 16 54 42](https://cloud.githubusercontent.com/assets/1197635/12454196/c46ef184-bf96-11e5-86ce-bcd752e7075f.png)

It seems that the action which is executed on "Refresh" is still "Add process" or most probably the last executed task.

tomasnorre:
Hmm.. Will have a look, just updated today in another project. Will have to check tomorrow.. 

Thanks for reporting.

PR #46, opened 2016-01-22
[BUGFIX] Fix the "Refresh" button in the processes information page f…

vdespa:
…or TYPO3 6 & 7 LTS.

The button would just re-execute the last task. If the last task was "Add process",
clicking refresh would add another process instead of refreshing the page.

Resolves: #45

### DIFF
--- a/view/process/class.tx_crawler_view_process_list.php
+++ b/view/process/class.tx_crawler_view_process_list.php
@@ -315,7 +315,7 @@ class tx_crawler_view_process_list {
 	 * @return string
 	 */
 	protected function getRefreshLink() {
-		return '<input onclick="window.location+=\'\';" type="button" style="padding:4px 4px 4px 20px; background-position: 3px 3px; background-image: url(\'' . $this->getIconPath() . 'arrow_refresh.png' . '\'); background-repeat: no-repeat;" value="' . $this->getLLLabel('LLL:EXT:crawler/modfunc1/locallang.xml:labels.refresh') . '" />';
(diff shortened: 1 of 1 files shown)


view/process/class.tx_crawler_view_process_list.php
```diff
@@ -315,7 +315,7 @@ protected function getRefreshIcon() {
 	 * @return string
 	 */
 	protected function getRefreshLink() {
-		return '<input onclick="window.location+=\'\';" type="button" style="padding:4px 4px 4px 20px; background-position: 3px 3px; background-image: url(\'' . $this->getIconPath() . 'arrow_refresh.png' . '\'); background-repeat: no-repeat;" value="' . $this->getLLLabel('LLL:EXT:crawler/modfunc1/locallang.xml:labels.refresh') . '" />';
+		return '<input onclick="window.location=\'' . \TYPO3\CMS\Backend\Utility\BackendUtility::getModuleUrl('web_info') . '&SET[crawlaction]=multiprocess' . '\';" type="button" style="padding:4px 4px 4px 20px; background-position: 3px 3px; background-image: url(\'' . $this->getIconPath() . 'arrow_refresh.png' . '\'); background-repeat: no-repeat;" value="' . $this->getLLLabel('LLL:EXT:crawler/modfunc1/locallang.xml:labels.refresh') . '" />';
```
skorpi: This is a huge line :-). What do you think about putting the refresh url into a variable?

tomasnorre: I think it's a good idea @skorpi .. Lets do that as part of some clean up...

PR #55, opened 2016-01-27
[TASK] Add refresh hooks

tomasnorre:
Add refresh hooks so that the refresh button also removes orphan and old processes.


tomasnorre:
Closed as it's replaced by pull request: #58

PR #58, opened 2016-02-02
[TASK] Add refresh hooks so that the refresh button …

tomasnorre:
It also removes orphan and old processes.

This replaces pull request: #55


modfunc1/class.tx_crawler_modfunc1.php
```diff
@@ -1231,13 +1233,31 @@ function selectorBox($optArray, $name, $value, $multiple)	{
 	}
 
 	/**
+	 * Activate hooks
+	 *
+	 * @return	void
+	 */
+	function runRefreshHooks() {
+		global $TYPO3_CONF_VARS;
```
srotsch: Please use $GLOBALS['TYPO3_CONF_VARS'] instead

Issue #8, opened 2015-09-01
Deleting old tx_crawler_processes entries ends up in slow query


tomasnorre:
This update query uses a lot of performance. Is there any way to improve this action?
To use UPDATE and set the process to deleted=1 leaves the table with a lot of old deleted entries.

$GLOBALS['TYPO3_DB']->exec_UPDATEquery( 
'tx_crawler_process', 
'active=0 
AND NOT EXISTS ( 
SELECT * FROM tx_crawler_queue 
WHERE tx_crawler_queue.process_id = tx_crawler_process.process_id 
AND tx_crawler_queue.exec_time = 0 
)', 
array( 
'deleted'=>'1' 
) 
);

We get this all the time:

Query 
1033 
Sending data 
UPDATE tx_crawler_process SET deleted = '1' WHERE active =0 AND NOT EXISTS (SEL

Is there a possibility to use DELETE instead of UPDATE, so the table doesn't fill up that much?

IMPORTANT: Please see dicussion/questions at: https://forge.typo3.org/issues/6792, furture discussion will continue here.

PR #76, opened 2016-03-17
[TASK] Deleting old tx_crawler_processes entries ends up in slow query

tomasnorre:
Resolves issue #8 

I've added the CLI_deleteProcessMarkedDeleted() to keep the process list smaller and therefor the slow query problem should be solved.

### DIFF
--- a/class.tx_crawler_lib.php
+++ b/class.tx_crawler_lib.php
@@ -2207,6 +2207,7 @@ class tx_crawler_lib {
 			}
 
 			$this->CLI_releaseProcesses($orphanProcesses, true); // maybe this should be somehow included into the current lock
+			$this->CLI_deleteProcessesMarkedDeleted();
 
 			$this->db->sql_query('COMMIT');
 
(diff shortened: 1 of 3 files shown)


class.tx_crawler_lib.php
```diff
@@ -2280,6 +2281,19 @@ function CLI_releaseProcesses($releaseIds, $withinLock=false) {
...
+	 */
+ 	public function CLI_deleteProcessMarkedDeleted($withinLock = false) {
+		if(!$withinLock) $this->db->sql_query('BEGIN');
+
+		$this->db->exec_DELETEquery('tx_crawler_process', 'deleted = 1');
+
+		if(!$withinLock) $this->db->sql_query('COMMIT');
+	}
+
+	/**
```
srotsch: As we only call this method at one place with $withinLock = true, the parameter might be removed completely.
The conditional code could be removed then, too.
Additionally, please rename to "CLI_deleteProcessesMarkedDeleted" (plural)

srotsch:
Resolves: #8

PR #96, opened 2016-07-06
replace deprecated function call in TCA file, add renderType for select fields

macjohnny:
this removes deprecation warnings in TYPO3 7.6


tomasnorre:
Thanks for your contribution @macjohnny

Issue #107, opened 2016-08-11
Deprecated Notices


maechler:
Some cli classes throw a deprecation notice using PHP 7 because they are not using __construct as constructor method name.

Core: Error handler (BE): PHP Runtime Deprecation Notice: Methods with the same name as their class will not be constructors in a future version of PHP; tx_crawler_cli_im has a deprecated constructor in typo3conf/ext/crawler/cli/class.tx_crawler_cli_im.php line 38

maechler:
Closing here because pull request has been merged: https://github.com/AOEpeople/crawler/pull/108

Thanks!

tomasnorre:
Thanks for your contribution.

PR #108, opened 2016-08-11
#107 Fixes deprecated notices

maechler:
Some cli classes throw a deprecation notice using PHP 7 because they are not using __construct as constructor method name.

Core: Error handler (BE): PHP Runtime Deprecation Notice: Methods with the same name as their class will not be constructors in a future version of PHP; tx_crawler_cli_im has a deprecated constructor in typo3conf/ext/crawler/cli/class.tx_crawler_cli_im.php line 38


tomasnorre:
Hi @maechler 

Thanks for your feedback/suggestion, i'll take a look at it.

PR #112, opened 2016-08-19
remove deprecated API call

macjohnny:


tomasnorre:
@macjohnny 

Thanks for you suggestion, but i'm trying to understand why you want to change this. 

The reason for using the 

`
$backRef->excludeIcon(
                $this->getContextMenuIcon()
            ),
`

Is that the use has the possibility to disable icons. Please let me know if i'm missing something in your suggestion.

macjohnny:
hey @tomasnorre 
the method is deprecated and not available in TYPO3 CMS 8 anymore.
See http://api.typo3.org/typo3cms/current/html/class_t_y_p_o3_1_1_c_m_s_1_1_backend_1_1_click_menu_1_1_click_menu.html#adc9776923e9198e90eb75c7207af7624

tomasnorre:
Hey @macjohnny 
Yeah I know, but it's still present in 7.6 and 6.2 so think i would suggest a version check to give the users the possibility to use "no icons" until the upgrade to 8.x

macjohnny:
i think this is a minor feature and therefore you could drop it, but if you want to implement a version check, go ahead.

tomasnorre:
It's a minor think agree.. But not the less it's an improvement and something that will fail "soonish".. So i think it's valid to implement the change. 

If you don't want to do it, i'll take care.. So please update the PR if so.: If not then i'll do next time i get to it.

macjohnny:
i think getting extensions ready for TYPO3 CMS 8 and further versions is crucial for a seamless and continuous upgrade cycle. moreover, the extension can be used for new projects with the latest typo3 version.
that said, maintaining some minor legacy features and adding code which should be removed again at some point is not that helpful.

tomasnorre:
True, i didn't consider it like that, I'll merge your PR, thanks for contributing.

Issue #118, opened 2016-09-06
Auto detect TypoScript Template


tomasnorre:
Implement logic that auto detects root TypoScript template for the for the page being crawled

PR #130, opened 2016-09-21
[TASK] Auto detect TypoScript Template

tomasnorre:
Solves issue #118 

The tests are running green locally, but as .travis.yml is not adjusted yet to run the namespaced Tests the Travis Build is not including the new test from here.

Would be cool if you could verify this locally @srotsch. 

From my tests it works as expected.

Update: For some reason the regular Functional tests failed, that i would have to look into. That didn't happen on the first commits.

### DIFF
--- a/Classes/Utility/TypoScriptUtility.php
+++ b/Classes/Utility/TypoScriptUtility.php
@@ -1,0 +1,60 @@
+<?php
+namespace AOE\Crawler\Utility;
+
+/***************************************************************
+ *  Copyright notice
+ *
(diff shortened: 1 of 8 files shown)


tomasnorre:
@srotsch I found the problem, but would like to discuss this. 

Current [WIP] commit is only proof of concept to verify that the tests are green when fixed.

tomasnorre:
Please do this where the crawling is happening and not in the constructor.. This will also remove the problem with the BE-module. 

Please add it to the readURL in tx_crawler_lib  and create early return for is_array() L:1153

PR #153, opened 2016-11-23
[BUGFIX] Fixed incorrect class name in ClassAliasMap

rvock:
The namespace for the CrawlerTask was incorrect.


tomasnorre:
@rvock Thanks, We will have a  look into it asap.

PR #192, opened 2017-01-30
[TASK] Write tests for tx_crawler_lib - Part III (#184)

tomasnorre:


Tests/Functional/CrawlerLibTest.php
```diff
@@ -58,6 +58,11 @@ public function setUp()
         $this->subject = $this->getAccessibleMock('\tx_crawler_lib', ['dummy']);
     }
 
+    public function teamDown()
```
srotsch: tear, not team :-) 

And if you only call the parent's method, you don't need it here at all.

tomasnorre: You're right, my bad.. Will remove.

PR #193, opened 2017-01-31
[BUGFIX] Starting a new cli process is not working with windows

domin1101:
It is currently not possible to start a cli crawler process in the backend when having a windows server.

I have found a pull request from the [old github mirror repo](https://github.com/TYPO3-extensions/crawler/pull/1), tested the changes and would now like to add them to the extension.


tomasnorre:
Thanks for your contribution.

PR #196, opened 2017-02-17
[BUGFIX] 8.x compatibility

thommyhh:
Make the extension work with TYPO3 CMS 8.x (8.6.0 especially):
* fix broken language label references
* remove invalid displayCondition "FALSE"
* replace IconUtility with IconFactory for 8.0.0 and above
* remove call to removed methods

### DIFF
--- a/Configuration/TCA/tx_crawler_configuration.php
+++ b/Configuration/TCA/tx_crawler_configuration.php
@@ -26,7 +26,7 @@ $GLOBALS['TCA']['tx_crawler_configuration'] = array(
     'columns' => array(
         'hidden' => array(
             'exclude' => 1,
-            'label' => 'LLL:EXT:lang/locallang_general.xml:LGL.hidden',
+            'label' => 'LLL:EXT:lang/Resources/Private/Language/locallang_general.xlf:LGL.hidden',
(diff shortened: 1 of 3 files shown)


thommyhh:
I introduced a version switch in two situations to also support 6.2 and 7, as they are set in composer.json.

Configuration/TCA/tx_crawler_configuration.php
```diff
@@ -26,7 +26,7 @@
     'columns' => array(
         'hidden' => array(
             'exclude' => 1,
-            'label' => 'LLL:EXT:lang/locallang_general.xml:LGL.hidden',
+            'label' => 'LLL:EXT:lang/Resources/Private/Language/locallang_general.xlf:LGL.hidden',
```
tomasnorre: But this is not backwards compatible as the language files are not move in earlier versions. 

We need to come up with a way to keep this compatible.

zdreicom: hi thommyh,  hi tomasnorre

i think is teh right way to put an Compatibility swich in Overrides/tx_crawler_configuration.php

bg

timo

PR #199, opened 2017-03-29
Show exception message

cweiske:
When an exception is thrown during crawling - e.g. because a hook throws one - it was caught without notice. This cost me 6 hours of debugging.

Now it is shown in debug output.


tomasnorre:
Thanks @cweiske , it is really useful information.

PR #204, opened 2017-07-10
[TASK] Use typo3/cms-core as dependency and replace typo3-ter/crawler

t3easy:


composer.json
```diff
@@ -18,14 +18,15 @@
     ],
     "minimum-stability": "dev",
     "require": {
-        "typo3/cms": ">=6.2.1,<8.9.99"
+        "typo3/cms-core": ">=6.2.1,<8.9.99"
     },
     "require-dev": {
```
tomasnorre: I think we need the typo3-cms in require-dev to be able to run the tests ?

tomasnorre: The error is not due to typo3-cms "missing" in require-dev.. 

I'll look into this and adjust the travis.yml

t3easy: No. You require it during runtime of the tests. You do it already. All tests for TYPO3 6.2 and 7.6 are green.

tomasnorre: Yeah i see.. I'll adjust the travis run. 

Excluding HHVM and allow failure in 8.* as it's not officially supported.. We have V8 branch for the one whom need it.

t3easy: For the large test range of 6.2 to 8.7 you could switch to nimut/testing-framework https://github.com/Nimut/testing-framework

tomasnorre: Thanks will check, didn't know about that one.

tomasnorre:
You change have been merge @t3easy. Thanks for helping out.

PR #211, opened 2017-10-09
[TASK] add post process signal slot also for use with crawler_im

samowitsch:
We need the post process signal slot also for use with the crawler_im cli task.


srotsch:
@samowitsch thanks for your contribution!

Issue #224, opened 2017-12-01
Drop Support for TYPO3 6.2


tomasnorre:
Remove versiom_compare() conditions

tomasnorre:
Pull Request can be created when 6.0.x branch is created.

tomasnorre:
As we can switch to fully composer based, should we drop the autoloading file as part of this issue to @srotsch ?

https://github.com/AOEpeople/crawler/blob/master/ext_autoload.php

PR #232, opened 2017-12-02
[TASK] Drop support for TYPO3 6.2 

tomasnorre:
Resolves: #224

### DIFF
--- a/Classes/Utility/IconUtility.php
+++ b/Classes/Utility/IconUtility.php
@@ -37,19 +37,13 @@ class IconUtility
     /**
      * Renders the HTML tag to show icons for a database record
      *
-     * Wrapper around core functionality to keep compatibility with TYPO3 6.2
-     *
      * @param $table
      * @param array $row
      * @return string
      */
     public static function getIconForRecord($table, array $row)
     {
-        if (version_compare(TYPO3_version,'7.0','<')) {
-            return \TYPO3\CMS\Backend\Utility\IconUtility::getSpriteIconForRecord($table, $row);
-        } else {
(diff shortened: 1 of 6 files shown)


Classes/Utility/IconUtility.php
```diff
@@ -37,19 +37,13 @@ class IconUtility
     /**
      * Renders the HTML tag to show icons for a database record
      *
-     * Wrapper around core functionality to keep compatibility with TYPO3 6.2
-     *
      * @param $table
      * @param array $row
      * @return string
      */
     public static function getIconForRecord($table, array $row)
```
tomasnorre: I kept this function to keep the implementation clean. 
Don't like to instantiate the iconFactory inside the classes when needed. 

I like this approach more.

modfunc1/class.tx_crawler_modfunc1.php
```diff
@@ -780,11 +780,7 @@ function drawLog_addRows($pageRow_setId, $titleString, $itemsPerPage=10) {
 				}
 
 				$setId = intval(\TYPO3\CMS\Core\Utility\GeneralUtility::_GP('setID'));
-
-				$refreshIcon = $GLOBALS['BACK_PATH'] . 'gfx/refresh_n.gif';
-				if (version_compare(TYPO3_version,'7.0','>=')) {
-					$refreshIcon = $GLOBALS['BACK_PATH'] . 'sysext/t3skin/extjs/images/grid/refresh.gif';
-				}
+                $refreshIcon = $GLOBALS['BACK_PATH'] . 'sysext/t3skin/extjs/images/grid/refresh.gif';
```
tomasnorre: Indentation problems.

But think this can be done with a prober PSR-2 converting..

PR #245, opened 2017-12-20
[BUGFIX] Added removed getIconForState + needed functions (#244)

### LINKED ISSUE #244: Add Process failed
Due to heavy cleanup the "Add process" failed.

tomasnorre:
Resolves #244

### DIFF
--- a/view/process/class.tx_crawler_view_process_list.php
+++ b/view/process/class.tx_crawler_view_process_list.php
@@ -305,6 +305,48 @@ class tx_crawler_view_process_list
     }
 
     /**
+     * Returns the state icon for the current job
+     *
+     * @param string $state
+     * @return string icon
+     */
+    protected function getIconForState($state)
+    {
+        switch ($state) {
+            case 'running':
+                $icon = 'bullet_orange';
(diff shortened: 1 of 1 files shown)


view/process/class.tx_crawler_view_process_list.php
```diff
@@ -304,6 +304,46 @@ protected function asMinutes($seconds)
...
+        }
+
+        return $this->getIcon($icon, $title);
+    }
+
+    /**
+     * Returns an imagetag for an icon
+     *
+     * @param string $icon
+     * @return string html tag for icon
```
srotsch: The PHPdoc annotation for the second parameter is missing

view/process/class.tx_crawler_view_process_list.php
```diff
@@ -304,6 +304,46 @@ protected function asMinutes($seconds)
...
+     */
+    protected function getIcon($icon, $title = '')
+    {
+        if (!empty($title)) {
+            $title = ' title="'.$title.'"';
+        }
+        return '<img src="'.$this->getIconPath().$icon.'.png" ' . $title . ' />';
```
srotsch: Whitespaces around the concatenation operators missing here and in line 342; we should add this to the php-cs-fixer config.

PR #268, opened 2018-01-22
[BUGFIX] Check if key exists before using count()

maximilian-walter:
The array-index is undefined if the checked event has no observers.
This leads to a warning with PHP 7.2, because count() must only be used
on arrays or objects which implements Countable.


tomasnorre:
Thanks for you contribution @maximilian-walter.

PR #270, opened 2018-01-24
[BUGFIX] Fix missing locallang labels for TYPO3 7.6

stucki:
Many core labels are gone in TYPO3 7.6 because the path for `locallang_core.php` and `locallang_general.php` was changed in EXT:crawler 6.0.

This fix will define an alternative location for the two files.

Note: I copied and adapted this fix from EXTnews. Credits for the solution go to @georgringer. Thank you!


tomasnorre:
Thanks to both @stucki and @georgringer.

Issue #279, opened 2018-04-05
[BUG] Can not create schedular tasks after installing crawler


r3h6:
After clicking on the plus-button for creating a new schedular task no backend form is displayed.
TYPO3 8.7 with composer.

r3h6:
After clicking on the "Add task" Button...
![a](https://user-images.githubusercontent.com/4550541/38990575-22aa1aec-43db-11e8-9bc5-ca6219ba40ce.png)

... I expect to get the form for creating a new task.
![b](https://user-images.githubusercontent.com/4550541/38990576-22dba3a0-43db-11e8-90f1-0a13ebf4a3e3.png)

... but after installing the "crawler" extension, I get this (note that there is a "Save" Button but no form):
![c](https://user-images.githubusercontent.com/4550541/38990577-230e5458-43db-11e8-9a34-3ad87738b49a.png)

tomasnorre:
Sorry but the screenshots doesn't help me a lot, I knew the steps, they don't bring me more information.. 

Do you get other errors than "no tasks defined yet" ?

r3h6:
The problem is in CrawlerQueueTaskAdditionalFieldProvider::getSelectedState (Line 149):
`$arraySize = count($configurationArray);`

$configurationArray is NULL...

tomasnorre:
@r3h6 Thanks for your PR.. Will have a look tomorrow.

As I cannot reproduce the Problem, I'll have to dig into it, before I merge the PR. 

I'll get back to you if I have any questions.

tomasnorre:
@r3h6 Can you please test if #284 fixes your problem. 

From the code, the $configurationArray cannot be empty anymore as I read it.. 
If it's working for you we have to fixes within one PR. 

If it is not working I'll look into to your PR again.

r3h6:
Issue "issue/190" doesn't solve the problem.

tomasnorre:
ok.. Thanks for letting us know. I'll take a look at your PR once again.

tomasnorre:
@r3h6 Can you please test with current master, we have added some fixes that should solve the problem.

r3h6:
With current master I get following exception:

Argument 1 passed to AOE\Crawler\Task\CrawlerQueueTaskAdditionalFieldProvider::getSelectedState() must be of the type array, null given, called in /var/www/web/typo3conf/ext/crawler/Classes/Task/CrawlerQueueTaskAdditionalFieldProvider.php on line 105 


[...]

PR #281, opened 2018-04-20
[BUGFIX] Check configurationArray for null

r3h6:
Resolves: #279


tomasnorre:
Rejected as fixed in #296

tomasnorre:
Thanks a lot for your input @r3h6 and for helping testing and giving feedback.

PR #293, opened 2018-05-07
[TASK] Testing/model process

tomasnorre:


Classes/Domain/Model/Process.php
```diff
@@ -43,12 +45,24 @@ class Process
...
+     */
+    protected $objectManager;
+
     /**
      * @param array $row
      */
     public function __construct($row = [])
     {
         $this->row = $row;
+        $this->objectManager = GeneralUtility::makeInstance(ObjectManager::class);
```
srotsch: Nitpick: I'd maybe make this a local variable only, or use a static getter, as it seems the ObjectManager is only used once in this class.

Tests/Functional/Domain/Model/ProcessTest.php
```diff
@@ -25,43 +25,90 @@
...
+
+    /**
+     * @var ObjectManager
+     */
+    protected $objectManger;
+
+    public function setUp()
+    {
+        parent::setUp();
+        $this->objectManger = GeneralUtility::makeInstance(ObjectManager::class);
```
srotsch: Same as above.

Issue #92, opened 2016-05-23
URL scheme 'http' is hardcoded


Idleworks:
In some cases it is necessary to call pages with https instead of http, especially on secured sites where port 80 is blocked.

URL scheme could be fetched from page's url_scheme property, baseUrl or newly created setting in extension manager.

tomasnorre:
Hi @Idleworks 

Thanks for your suggestion it makes perfectly sense to adapt this.

ogrosko:
Hey
Has this feature something in common with my problem:
When I run Crawler queue task from typo3 BE scheduler module my queued url is 
`https://mydomain.com/index.php?id=1` and url is crawled correctly

but wehn I run crawler queue task from cli `./typo3/cli_dispatch.phpsh scheduler -i 3 -f
` (3 is the id of crawler task in scheduler) queued url is `http://./index.php?id=1` and after crawling url I get an error.

Where is the problem. Missed I some configuration for ssl connection?
Thanks for any advice

ogrosko:
Ok I make some investigation:

1. `displayCond` config for `base_url` field in `/typo3conf/ext/crawler/Configuration/TCA/tx_crawler_configuration.php` doesn't work. I have to change it from 
` 'displayCond' => 'FIELD:sys_domain_base_url:=:0',`
to
` 'displayCond' => 'FIELD:sys_domain_base_url:<=:0',`
now it works and when I use this field and set my domain to `https://mydomain.com/`

But it still not working when you use `sys_domain_base_url` (Get Baseurl from Domainrecord) field
It doesn't accept https. And this is object of this task ...

tomasnorre:
Hi @ogrosko 

Thanks for your elaboration we will take a look at it.

CDRO:
Hey people of AOE,

thanks for taking care of the TYPO3 crawler :-)

But it seems that in the latest release, it's impossible again to work with websites using https? Or am I missing something in my config?

Best Regards
Tizian

tomasnorre:
Hi @CDRO 

It should be possible.
Can you please elaborate a little more ?

We are doing it in other project, our configuration is using direct_requests.

CDRO:
My setup is are as follows:
- PHP 7
- TYPO3 7.6
- Linux Ubuntu with Apache 2.4 or Windows with Apache 2.4/IIS 7-8

[...]

PR #299, opened 2018-05-08
[FEATURE] Adds ForceSsl to Crawler Configuration (#92)

tomasnorre:


Classes/Controller/CrawlerController.php
```diff
@@ -654,20 +654,24 @@ protected function getUrlsForPageId($id, $forceSsl = false)
                                 $TSparserObject = GeneralUtility::makeInstance('TYPO3\CMS\Core\TypoScript\Parser\TypoScriptParser');
                                 $TSparserObject->parse($configurationRecord['processing_instruction_parameters_ts']);
 
+                                // Check Crawler Configurations forceSsl if false use the Page url_scheme.
+                                $crawlerWithSsl = $configurationRecord['force_ssl'] ?: $forceSsl;
```
srotsch: In general, I think this is a feasible solution.

The only downside I can image: this only allows a "positive" override: given a couple of pages with URL scheme forced to HTTP and some others forced to HTTPS, this setting would allow crawling all pages _with_ SSL. But it's not possible to enforce crawling all pages _without_ SSL.

In case we want to support "negative" enforcement too, maybe we can make this field a dropdown with e.g. -1, 0, 1, and:
- "-1" enforces HTTP
- 0 keeps each page's URL scheme
- "1" enforces HTTPS

Don't mind about the actual values, but I hope you get the idea.

Classes/Controller/CrawlerController.php
```diff
@@ -654,20 +654,24 @@ protected function getUrlsForPageId($id, $forceSsl = false)
...
-                                        $forceSsl
+                                        $crawlerWithSsl
                                     ),
                                     'realurl' => $configurationRecord['realurl'],
                                     'cHash' => $configurationRecord['chash'],
                                     'userGroups' => $configurationRecord['fegroups'],
                                     'exclude' => $configurationRecord['exclude'],
                                     'rootTemplatePid' => (int) $configurationRecord['root_template_pid'],
                                     'key' => $key,
+                                    'force_ssl' => $configurationRecord['force_ssl'],
```
srotsch: Not quite sure where this is used, but if it's somewhere visible to the user it should maybe contain what we are _really_ crawling the page with. With the current code, maybe this should be the value of $crawlerWithSsl.

Classes/Controller/CrawlerController.php
tomasnorre: I think this is a good idea.. Will look into this.

PR #303, opened 2018-05-16
[TASK] Re-adds type hint (#10)

### LINKED ISSUE #10: Re-add type hint to tx_crawler_hooks_staticFileCacheCreateUri::initialize
The type hint for $parent has been removed as a bugfix with #66545. With switching to namespaces, we should add it again:

@public function initialize(array $parameters, SFC\NcStaticfilecache\StaticFileCache $parent) {

Migrated from: https://forge.typo3.org/issues/68727

tomasnorre:
Resolves #10

### DIFF
--- a/Classes/Hooks/StaticFileCacheCreateUriHook.php
+++ b/Classes/Hooks/StaticFileCacheCreateUriHook.php
@@ -38,10 +38,11 @@ class StaticFileCacheCreateUriHook
      * Initializes the variables before starting the processing.
      *
      * @param array $parameters The parameters used in this hook
-     * @param $parent The calling parent object
-     * @return void
+     * @param \SFC\NcStaticfilecache\StaticFileCache $parent The calling parent object
+     *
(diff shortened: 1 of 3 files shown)


Classes/Hooks/StaticFileCacheCreateUriHook.php
```diff
@@ -38,10 +38,11 @@ class StaticFileCacheCreateUriHook
      * Initializes the variables before starting the processing.
      *
      * @param array $parameters The parameters used in this hook
-     * @param $parent The calling parent object
-     * @return void
+     * @param \SFC\NcStaticfilecache\StaticFileCache $parent The calling parent object
+     *
+     * @throws \Exception
```
srotsch: You accidentally removed the return annotation

tomasnorre: Ups.. Will update

PR #309, opened 2018-07-06
Support for PageTSConfig with processing instructions only

nicodh:
This is not really tested and needs thorough review since I did not analyze the whole crawler implementation!


tomasnorre:
Thanks, Will have a look at soon as possible.

Classes/Controller/CrawlerController.php
```diff
@@ -618,14 +617,47 @@ protected function getUrlsForPageId($id, $forceSsl = false)
...
+
+                                    // add trailing slash if not present
+                                if (!empty($subCfg['baseUrl']) && substr($subCfg['baseUrl'], -1) != '/') {
+                                    $subCfg['baseUrl'] .= '/';
+                                }
+
+                                // Explode, process etc.:
+                                $res[$key] = [];
+                                $res[$key]['subCfg'] = $subCfg;
+                                $res[$key]['paramParsed'] = [];
```
tomasnorre: I'm been reading through the code several times, and trying to understand why you are setting the $res[$key]['paramPased'] = [] (empty array).

And not like it is in the first part with 
`$res[$key]['paramParsed'] = $this->parseParams($values);` you are also ignoring the `$res[$key]['paramExpanded']` 

Can you explain to me why?

nicodh: Well I think the 2 possible scenarios are:
A. a setting like
key = params
key {
    instructions
}

B. or only 
key {
  instructions
}

A. produces 2 keys, on with and one without dot
B results in only one key with dot

So with B there are no params to be parsed

I made a mistake which resulted in overriding the previous key which I fixed with my 2nd commit

tomasnorre:
Hi @nicodh 
Could you please rebase you changes, then I'll make sure it get tested. 
Thanks, Happy New Year.

nicodh:
Done!

tomasnorre:
You were faster than I expected :)

I'm on vacation for the next 14 days, will see if I can find the time to check before leaving, otherwise I'll check when I get back.

nicodh:
No hurry ;-)
To be honest I'm not really aware anymore about the details and solved the problem with another workaround so far...

tomasnorre:
That's the downside of slow responses.. I'm not really 100% i grasp it anymore either, but will check if i get it.

tomasnorre:
When I add following configuration. 

```
tx_crawler.crawlerCfg.paramSets.tomasnorre.procInstrFilter = tx_indexedsearch_reindex
```

Then I get the error: 
```
Argument 2 passed to TYPO3\CMS\Core\Utility\GeneralUtility::implodeArrayForUrl() must be of the type array, null given, called in /var/www/congstar-congo/devbox/htdocs/typo3conf/ext/crawler/Classes/Controller/CrawlerController.php on line 2081
```

I don't see at first glance what the problem is, would need to trigger the debugging on this.

tomasnorre:
Hi @nicodh 

We are about to release the 9.0.0 version, and switched the branches around. This closed this PR automatically. 

Can you please recreate it with reference to `TYPO3_7` instead.
Thanks

tomasnorre:
Hi @nicodh 

Could you please rebase this again master, then I'll have a look at it again. Sorry for the extremely long waiting time.

tomasnorre:
Still relevant.

tomasnorre:
Please rebase @nicodh

tomasnorre:
Closed in favour of: https://github.com/AOEpeople/crawler/pull/737

Issue #311, opened 2018-08-02
[BUG] TSConfig Crawler configurations still not working correctly


stuartmcfarlane:
`getUrlsForPageId()` incorrectly calls `parseParams()` with the `key.` array instead of the `key`'s value causing `parseParams()` to do nothing

tomasnorre:
Closed as fixed in PR #312

PR #312, opened 2018-08-02
[BUGFIX] paramSets correctly parsed (#311)

stuartmcfarlane:


tomasnorre:
Thanks for you PR, this PR is already touching some problems in this area. 
https://github.com/AOEpeople/crawler/pull/309

I'll look into this.

PR #320, opened 2018-11-06
[BUGFIX] Replace old class name for LanguageService

tbal:
Class 'language' is deprecated since 6.0 and removed with 7.0.
This patch replaces the old 'language' class name usage
with LanguageService.


tomasnorre:
Thanks.

Issue #314, opened 2018-09-26
TCA Label ofr hidden checkbox needs to be changed


tomasnorre:
TCA has a missing label for hidden checkbox, has to be changed to:
'label' => 'LLL:EXT:lang/Resources/Private/Language/locallang_general.xlf:LGL.hidden',

TYPO3: 8.7

tomasnorre:
Resolved with https://github.com/AOEpeople/crawler/commit/6115106de62b1e0af6462b9602194447bb3ced47

PR #321, opened 2018-11-11
[BUGFIX] Changed TCA Label for hidden checkbox

ayacoo:
Related #314


tomasnorre:
We are still supporting TYPO3 7.6, so this is not a valid fix. 

see: 
https://github.com/TYPO3/TYPO3.CMS/tree/TYPO3_7-6/typo3/sysext/lang/Resources/Private/Language

The file you are referring to doesn't exists in TYPO3 7.6, but I'm not 100% if this will work anyway because of the mapping in ext_tables.php

https://github.com/AOEpeople/crawler/blob/master/ext_tables.php

tomasnorre:
Ping @ayacoo

ayacoo:
Hi @tomasnorre 

Sorry for the late response. I checked with an old T3 7.6.x and it works because of the version mapping.

PR #324, opened 2018-11-15
[CLEANUP] Remove obsolete TCA feInterface config

tbal:
feInterface-section in TCA configuration is deprecated since 6.0 and removed with 6.2.

See also: https://docs.typo3.org/typo3cms/TCAReference/6.0/Reference/Feinterface/Index.html


tomasnorre:
Thanks for you contribution @tbal

PR #325, opened 2018-11-19
 [BUGFIX] Avoid warning if $q_entry['result_data'] is empty

revoltek-daniel:
If $q_entry['result_data'] is empty it will trigger an PHP Warning on unset.


tomasnorre:
Thanks @revoltek-daniel

PR #328, opened 2018-11-21
[FEATURE] Use Extbase Domain/Model + Repository #323

### LINKED ISSUE #323: [FEATURE] Convert Domain/Model + Repositories to TYPO3 AbstractEntity & Repository
We are currently having a Domain Model + Repositories based on an older approach to Model and Repository. We would like to have this converted into Domain/Model that extends the AbstractEntity and the Domain/Repository extending the Persistence/Repository, so that we can benefit from the functionalities that comes build-in with TYPO3. 

This will enables us to remove quite an amount of code. 

We are currently having following Domain/Model:
* Process
* ProcessCollection
* Queue
* Reason

We have following Domain/Repository
* AbstractRepository
* ProcessRepository
* QueueRepository

What is missing is a Domain/Model for `Configuration` and a related Domain/Repository of course.

**Import** 
If the database structure changes, it needs to be marked as Breaking Change, and the class.ext_update.php needs to be provided.

tomasnorre:
Resolves #323


tomasnorre:
I've still need some manual testing, but the review can be started.

bmack:
👍 looks good!

Issue #307, opened 2018-06-11
PHP Warning in CrawlerController in CLI mode


echavaillaz:
Hi,

When trying to run a Crawler task (For example "Crawler queue"), I got an  error : 

`Execution of task "Crawler queue (crawler)" failed with the following message: PHP Warning: explode() expects parameter 2 to be string, array given in /var/www/www.domain.tld/httpsdocs/typo3conf/ext/crawler/Classes/Controller/CrawlerController.php line 821`

The problem comes from the line 618 `$res[$key]['paramParsed'] = $this->parseParams($values);`

Thanks,
Eric

tomasnorre:
Thanks for the information @echavaillaz 

I'll have a look at it. 

Please let me know which TYPO3 version and Crawler version.

echavaillaz:
Last TER version for Crawler and TYPO3 7.6.28.
Thanks!

tomasnorre:
Thanks for the additional information @echavaillaz , will look into this, please be patient, I'll leave for holidays soon, so I cannot promise a response right away.

echavaillaz:
No problem, I have found a way to bypass the warning yet. But I think it is a real bug you have to resolve here.

Nice holidays!

tomasnorre:
You're more than welcome to provide a bugfix for this. Pull request are happily reviewed :)

tomasnorre:
@echavaillaz Can you please paste me the exact command you are running from CLI?

echavaillaz:
My solution is naive so I don't think a pull request is the right way now. But I could investigate a little more if you need my help.

For the command, there is several way to have the warning, for example, simply when running the task from the backend module or with A CRON command on the server.

Thanks

tomasnorre:
ok. You wrote CLI mode in the title, so expected it to be from CLI.

tomasnorre:
It's logic that we have a conflict here: 

`
 if (is_array($values)) {
                        ....
                        code-removed
                        ....
                            $res[$key]['paramParsed'] = $this->parseParams($values);
`

`
/**
     * Parse GET vars of input Query into array with key=>value pairs
     *
     * @param string $inputQuery Input query string
     * @return array
     */
    public function parseParams($inputQuery)
    {
        // Extract all GET parameters into an ARRAY:
        $paramKeyValues = [];
        $GETparams = explode('&', $inputQuery);
       ....
      code-removed
       .... 
    }
`

[...]

PR #329, opened 2018-12-02
[BUGFIX] https://github.com/AOEpeople/crawler/issues/307

daCyberpunk:
Needed this urgently so That commit shoud fix the issue.

### DIFF
--- a/Classes/Controller/CrawlerController.php
+++ b/Classes/Controller/CrawlerController.php
@@ -599,7 +599,7 @@ class CrawlerController
                     // Sub configuration for a single configuration string:
                     $subCfg = (array)$crawlerCfg['paramSets.'][$key . '.'];
                     $subCfg['key'] = $key;
-                    $res[$key] = $res[$key] ?? [];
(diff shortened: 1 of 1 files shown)


daCyberpunk:
boah php 5..... wann war das nochmal? :-D

Classes/Controller/CrawlerController.php
```diff
@@ -589,17 +589,24 @@ protected function getUrlsForPageId($id, $forceSsl = false)
 
         $res = [];
 
-        if (is_array($pageTSconfig) && is_array($pageTSconfig['tx_crawler.']['crawlerCfg.'])) {
+        if (\is_array($pageTSconfig) && \is_array($pageTSconfig['tx_crawler.']['crawlerCfg.'])) {
```
tomasnorre: Why the \ before is_array() it is a function and not a Class?

ayacoo: https://decisions.typo3.org/t/should-we-use-fully-qualified-function-fqn-calls-like-is-array-in-the-core/392

Issue #252, opened 2017-12-21
[TASK] Deprecate custom event dispatcher


srotsch:
* add signals and use in parallel
* deprecate the custom event dispatcher
* fix stupid typo in existing signals

tomasnorre:
Resolved in #332

PR #332, opened 2018-12-13
[FEATURE] Deprecates EventDispatcher + Adds Signal Slots (#252)

tomasnorre:


kaystrobach:
Looks good from reading, wouldn’t it make sense to wrap the old signal with a function to make it possible to have an entry in the deprecation log?

tomasnorre:
Thanks @kaystrobach 

That makes perfectly good sense. Will do that.

bmack:
👍

tomasnorre:
@kaystrobach and @bmack 
What do you think about adding deprecations to the class
https://github.com/AOEpeople/crawler/blob/issue/252/Classes/Event/EventDispatcher.php

Then I don't need to wrapper it in functions like Kay suggested? It's sort of already wrapped.

kaystrobach:
Valid point

tomasnorre:
Is it enough to deprecate the constructor? or should I deprecate every single function, isn't that many so no a problem, just wanna know what you suggests.

kaystrobach:
I think the constructor is enough if the medsage is clear, which means, the whole object should not be used.

PR #334, opened 2019-02-18
[CLEANUP] Fix problems with PHP 7.2

tomasnorre:
crawler-php72


kaystrobach:
Looks good from reading

PR #335, opened 2019-03-04
CommandController implementations

tomasnorre:


Classes/Command/CrawlerCommandController.php
```diff
@@ -0,0 +1,266 @@
...
+     * @param int $number Specifies how many items are put in the queue per minute. Only valid for output mode "queue"
+     * @param string $conf A comma separated list of crawler configurations
+     */
+    public function buildQueueCommand($startpage = 0, $depth = 0, $mode = '', $number = 0, $conf = '')
+    {
+
+        /** @var CrawlerController $crawlerController */
+        $crawlerController = $this->objectManager->get(CrawlerController::class);
+
+        // Force user to admin state and set workspace to "Live":
```
felixsemmler: Can we remove these 3 lines?

Classes/Command/CrawlerCommandController.php
```diff
@@ -0,0 +1,266 @@
...
+//        $this->backendUser->user['admin'] = 1;
+//        $this->backendUser->setWorkspace(0);
+
+        if ($mode === 'exec') {
+            $crawlerController->registerQueueEntriesInternallyOnly = true;
+        }
+
+        if (1===2) {
```
felixsemmler: Is 1 === 2 another result if crawler is called over cli instead of TYPO3 BE?

Classes/Controller/CrawlerController.php
```diff
@@ -2211,12 +2213,14 @@ public function CLI_main_flush()
     /**
      * Obtains configuration keys from the CLI arguments
      *
-     * @param  QueueCommandLineController $cliObj    Command line object
-     * @return mixed                        Array of keys or null if no keys found
+     * @param $conf string
```
felixsemmler: Order should be type and then variable name:
> @param string $conf

composer.json
```diff
@@ -17,11 +17,12 @@
...
     "minimum-stability": "dev",
     "require": {
-        "typo3/cms-core": ">=7.6.0,<8.7.99"
+        "typo3/cms-core": ">=7.6.0,<8.7.99",
+        "helhum/typo3-console": "^4.0"
     },
     "require-dev": {
         "typo3/cms": "^7.6",
-        "nimut/testing-framework": "^2.0"
+        "nimut/testing-framework": "*"
```
felixsemmler: Is the "*" constraint for travis?

Classes/Command/CrawlerCommandController.php
tomasnorre: Yes, I forgot about them.

Classes/Command/CrawlerCommandController.php
tomasnorre: I think it's test data I forgot to remove, will check.

Classes/Controller/CrawlerController.php
tomasnorre: Correct, will fix.

composer.json
tomasnorre: Yes

Classes/Command/CrawlerCommandController.php
```diff
@@ -0,0 +1,262 @@
...
+        switch ($mode) {
+            case 'all':
+                $crawlerController->getLogEntriesForPageId($pageId, '', true, $fullFlush);
+                break;
+            case 'finished':
+            case 'pending':
+                $crawlerController->getLogEntriesForPageId($pageId, $mode, true, $fullFlush);
+                break;
+            default:
+                $this->outputLine("<info>No matching parameters found." . chr(10) . "Try 'typo3cms help crawler:flushqueue' to see your options</info>");
```
srotsch: I'd rather use PHP_EOL, if I remember correctly chr(10) is Linux only.
And maybe use single quotes for the string and the double quotes inside?

Classes/Command/CrawlerCommandController.php
```diff
@@ -0,0 +1,262 @@
...
+    public function buildQueueCommand($startpage = 0, $depth = 0, $mode = '', $number = 0, $conf = '')
+    {
+
+        /** @var CrawlerController $crawlerController */
+        $crawlerController = $this->objectManager->get(CrawlerController::class);
+
+        if ($mode === 'exec') {
+            $crawlerController->registerQueueEntriesInternallyOnly = true;
+        }
+
```
srotsch: Maybe get rid of some linebreaks? ;-)

Classes/Command/CrawlerCommandController.php
```diff
@@ -0,0 +1,262 @@
...
+
+
+        if (defined('TYPO3_MODE') && 'BE' === TYPO3_MODE) {
+            // Crawler is called over TYPO3 BE
+            $pageId = MathUtility::forceIntegerInRange(1, 0);
```
srotsch: Hmmm… why not setting the value directly? There's not much to force here.

Classes/Command/CrawlerCommandController.php
```diff
@@ -0,0 +1,262 @@
...
+                } else {
+                    $this->outputLine("<errror>Error checking Crawler Result:  " . substr(preg_replace('/\s+/', ' ', strip_tags($result['content'])), 0, 30000) . '...' . chr(10) . "</errror>" . chr(10));
+                }
+            }
+        } elseif ($mode === 'queue') {
+            $this->outputLine("<info>Putting " . count($crawlerController->urlList) . " entries in queue:</info>" . chr(10));
+            $this->outputLine("<info>" . implode(chr(10), $crawlerController->urlList) . "</info>" . chr(10));
+        } else {
+            $this->outputLine("<info>" . count($crawlerController->urlList) . " entries found for processing. (Use --mode to decide action):</info>" . chr(10));
+            $this->outputLine("<info>" . implode(chr(10), $crawlerController->urlList) . "</info>" . chr(10));
```
srotsch: I assume this code already existed… damn, this is ugly :-P

Classes/Command/CrawlerCommandController.php
```diff
@@ -0,0 +1,262 @@
...
+     *
+     * @return int
+     */
+    public function crawlQueueCommand($amount = 0, $sleeptime = 0, $sleepafter = 0)
+    {
+        $result = self::CLI_STATUS_NOTHING_PROCCESSED;
+
+        /** @var CrawlerController $crawlerController */
+        $crawlerController = $this->objectManager->get(CrawlerController::class);
+        $queueRepository = $this->objectManager->get(QueueRepository::class);
```
srotsch: For consistency you might want to add a type hint here as well.

PR #346, opened 2019-04-17
Rewrite TYPO3_DB to DBAL

CDRO:
I've spent some work time rewriting the Crawler to use `DBAL` instead of `TYPO3_DB`.

This is still a WIP and **has not been tested** but it covers all rewrites that are needed.

There is **one** query that has been rewritten where I was not 100% sure that this is indeed right, **but** I left the original code for direct comparison.


tomasnorre:
Thanks for you PR @CDRO 
But as we are still supporting TYPO3 v7 we cannot merge this yet.

tomasnorre:
I would be happy to work with you on this when we can drop support for TYPO3 v7.

CDRO:
It would be really cool if a pre-release could be scheduled for support of TYPO3 9 or some sort of release plan for TYPO3 9 would be nice too, have you any thought already on this?

tomasnorre:
Ja, I have started a re-write of the extension, but there are also a WIP branch for Version 9. 
https://github.com/AOEpeople/crawler/tree/typo3v9

tomasnorre:
But a timeline I don't have.. Working on this only during my spare time, so time is limited.

CDRO:
I've seen that you started a new branch, that's why my PR goes into it :-)

image

Do you have a list of needs that you'd like to put into the rewrite before releasing it?

tomasnorre:
Sorry didn't see that.. 

No unfortunately not.. but that would probably be very helpful, not only for me , but also for people wanting to help.. 

I'll try to make issues for all needed, it also makes PR and Reviews easier.

Issue #347, opened 2019-04-23
Make backend module compatible with TYPO3 9


CDRO:
The crawler backend module heavily relies on old backend API for document formatting.

I'll try to work on this and make an appropriate PR asap.

PR #348, opened 2019-04-23
Make module compatible with TYPO3 9

CDRO:
This solves parts of issue #347 (it has some ugly hacks though).

In order to get a quickly running solution, I implemented the section functions from the `DocumentTemplate` class into the `BackendModule` class.

I'll probably work on this PR since I don't thing, that the fix for the missing configurations from the backend should be placed here and that they make much more sense in #346

### DIFF
--- a/Classes/Backend/BackendModule.php
+++ b/Classes/Backend/BackendModule.php
@@ -209,8 +209,6 @@ class BackendModule extends AbstractFunctionModule
     {
         global $LANG, $BACK_PATH;
 
-        //$this->incLocalLang();
-
         $this->extensionSettings = ExtensionSettingUtility::loadExtensionSettings();
         if (empty($this->pObj->MOD_SETTINGS['processListMode'])) {
             $this->pObj->MOD_SETTINGS['processListMode'] = 'simple';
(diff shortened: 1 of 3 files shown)


Classes/Backend/BackendModule.php
```diff
@@ -204,7 +209,7 @@ public function main()
     {
         global $LANG, $BACK_PATH;
 
-        $this->incLocalLang();
+        //$this->incLocalLang();
```
tomasnorre: Please remove when commented out anyway :)

Classes/Backend/BackendModule.php
```diff
@@ -1016,11 +1174,11 @@ protected function makeCrawlerProcessableChecks()
     {
         global $LANG;
 
-        if ($this->isCrawlerUserAvailable() === false) {
+        /*if ($this->isCrawlerUserAvailable() === false) {
```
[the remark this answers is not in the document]
CDRO: Done in 72fde48

Classes/Backend/BackendModule.php
CDRO: Done in 72fde48

PR #357, opened 2019-05-17
[BUGFIX] Resolves the bin-dir from composer.json

### LINKED ISSUE #355: [BUG] CLI-Path in info module is wrong
The path to the CLI command in the module `Info` is wrong.
Example:
`/usr/local/bin/php_cli /html/releases/current/publictypo3cms crawler:crawlqueue`

1. The slash between the folder and `typo3cms` is missing
2. The binary is `/html/releases/current/vendor/bin/typo3cms` in composer mode by default not in the public path

I would use the path vendor/bin in composer mode and  typo3conf/ext/ in not.
Or remove the CLI-Path completely as the path to typo3cms depends on many different configurations.

tomasnorre:
Resolves #355


Classes/Service/ProcessService.php
```diff
@@ -221,11 +221,13 @@ public function startProcess()
      */
     public function getCrawlerCliPath()
     {
+        $composerRootDir = getenv('TYPO3_PATH_COMPOSER_ROOT') . '/';
+        $jsonDecoded = json_decode(file_get_contents($composerRootDir . 'composer.json'), true);
+        $binDir = $jsonDecoded['config']['bin-dir'] ?? 'vendor/bin';
```
tomasnorre: This is not compatible with PHP 5.6 would need to change this.

PR #380, opened 2019-05-29
[TASK] Update composer dependencies

tomasnorre:


composer.json
```diff
@@ -17,15 +17,21 @@
     ],
     "require": {
         "php": ">= 5.6, < 7.3",
```
aoekrz: Can we require php 7.0 as minimum?

PR #399, opened 2019-08-05
[TASK] Adapt version constraints

CDRO:


tomasnorre:
Same here, need to wait for next release.

PR #414, opened 2019-08-21
[TASK] Rework backend module to use Fluid

bmack:
The Backend module is rewritten using Fluid
and primitive action methods to render most
output on the Info Module => Crawler.

Base styling is available, but some minor changes
within the ProcessOverview need some improvements.

The old template functionality (incl pagination,
as this was not used anymore anyways) for ProcessOverview
is removed. Additional icons and css files are
removed as well, and TYPO3 Backend styling is used.

Usages of `global $LANG` is removed, most
rendering except for dropdowns is now built in fluid.

Rendering of nested columns within the Log view is still
compiled in the controller.

Controller does not inherit from AbstractFunctionModule
anymore as this class was deprecated in TYPO3 v9.


Classes/Backend/BackendModule.php
```diff
@@ -655,192 +423,132 @@ public function icons($type, $styleAttribValue = '')
...
+                    ->from('tx_crawler_queue')
+                    ->select('*')
+                    ->where(
+                        $this->queryBuilder->expr()->eq('qid', $this->queryBuilder->createNamedParameter($queueId))
+                    )
+                    ->execute()
+                    ->fetch();
+
+                // Explode values
+                $q_entry['parameters'] = unserialize($q_entry['parameters']);
```
tstahn: Do we have an opinion on passing `['allowed_classes' => false]` as a second paramater to `unserialize` in order to prevent potential PHP Object Injection? This is AFAIR _technically_ only an issue when user inputs come into play, but better safe than sorry?

Classes/Backend/BackendModule.php
```diff
@@ -655,192 +423,132 @@ public function icons($type, $styleAttribValue = '')
...
-                        'row' => $pageinfo,
-                        'HTML' => $HTML
-                    ];
-
-                    // Get branch beneath:
-                    if ($this->pObj->MOD_SETTINGS['depth']) {
-                        $tree->getTree($this->pObj->id, $this->pObj->MOD_SETTINGS['depth'], '');
-                    }
+                $this->view->assign('queueStatus', $resStatus);
+                $this->view->assign('queueDetails', DebugUtility::viewArray($q_entry));
```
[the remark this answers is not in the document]
bmack: that's how it was built before - "show details of all data as the debug utility does it" - just some copy/paste here. I think we should keep that as is for now.

Classes/Backend/BackendModule.php
bmack: yeah, thought about that too. there are so many `unserialize()` calls. Unfortunately I don't know what is stored in these fields (objects? i don't know). In general I suggest to move to json_encode/decode for all these things. Make that breaking and flush your queue, basically.

This would need adaption in the crawling process as well. Should we do this separately?

Classes/Backend/BackendModule.php
```diff
@@ -130,158 +135,141 @@ class BackendModule extends AbstractFunctionModule
...
+    /**
+     * Called by the InfoModuleController
+     * @param InfoModuleController $pObj
+     */
+    public function init(InfoModuleController $pObj): void
+    {
+        $this->pObj = $pObj;
+        $this->id = (int)GeneralUtility::_GP('id');
+        // Setting MOD_MENU items as we need them for logging:
+        $this->pObj->MOD_MENU = array_merge($this->pObj->MOD_MENU, $this->modMenu());
```
tomasnorre: Isn't the $this->pObj->MOD_MENU Deprecated? 
https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/9.4/Deprecation-85196-ProtectSetupModuleController.html

Classes/Backend/BackendModule.php
tomasnorre: We can do this later on, but could you please add an new issue for it, so that We/I don't forget about it.

Classes/Backend/BackendModule.php
```diff
@@ -1344,39 +978,49 @@ protected function addWarningMessage($message)
...
+            if (is_object($hookObj)) {
+                $hookObj->crawler_init($crawlerLib);
             }
         }
     }
+
+    protected function initializeView(): void
+    {
+        $view = GeneralUtility::makeInstance(StandaloneView::class);
+        $view->setLayoutRootPaths(['EXT:crawler/Resources/Private/Layouts']);
```
tomasnorre: Is this how it's done now a days? Or only because we have no TypoScript doing it for us?

tomasnorre:
@bmack You have also forgotten to add you to the Contributers.md

Classes/Backend/BackendModule.php
bmack: You don't want Extbase (which we don't have) to boot up TypoScript to allow extension devs to change your template... at least not for crawler ;).

We do this in a few other places in core like this (e.g. ToolbarItems etc).

Classes/Backend/BackendModule.php
bmack: nope. that's the setup module. Info module is all fine.

tomasnorre: So in our report is a false positive ?
https://report.typo3-crawler.com/

bmack: looks like it. good to know

PR #418, opened 2019-08-21
[BUGFIX] Changed the QueryBuilder for the Delete,Select and Update to avoid conflicts

tomasnorre:
This is not done yet, as we still cannot add Processes from backend.

But part of the problem is solved, there CommandController isn't throwing exceptions anymore.

### DIFF
--- a/Classes/Controller/CrawlerController.php
+++ b/Classes/Controller/CrawlerController.php
@@ -2228,7 +2228,6 @@ class CrawlerController implements LoggerAwareInterface
      */
     public function CLI_run($countInARun, $sleepTime, $sleepAfterFinish)
     {
-        $queryBuilder = GeneralUtility::makeInstance(ConnectionPool::class)->getQueryBuilderForTable($this->tableName);
         $result = 0;
         $counter = 0;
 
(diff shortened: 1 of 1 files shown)


Classes/Controller/CrawlerController.php
```diff
@@ -2277,12 +2275,14 @@ public function CLI_run($countInARun, $sleepTime, $sleepAfterFinish)
         if (intval($this->extensionSettings['purgeQueueDays']) > 0) {
             $purgeDate = $this->getCurrentTime() - 24 * 60 * 60 * intval($this->extensionSettings['purgeQueueDays']);
 
-            $del = $queryBuilder
+            $queryBuilderDelete = GeneralUtility::makeInstance(ConnectionPool::class)->getQueryBuilderForTable($this->tableName);
```
tstahn: Certainly nothing for now, but at some point we might want to split these database operations into separate methods. All this `$querBuilder*` juggling with its selections, deletions, updates, ... is pretty confusing.

tomasnorre: Totally agree. This should be functions called instead, to keep it cleaner.

bmack:
I think we should get it in as it is, however, there are places where we could use the Connection directly (e.g. UPDATE) instead of using another QueryBuilder instance.

tomasnorre:
Will rebase and merge

Classes/Controller/CrawlerController.php
```diff
@@ -2239,27 +2238,31 @@ public function CLI_run($countInARun, $sleepTime, $sleepAfterFinish)
...
+            $del = $queryBuilderDelete
                 ->delete($this->tableName)
                 ->where(
                     'exec_time != 0 AND exec_time < ' . $purgeDate
-                );
-            if (false == $del) {
-                $this->logger->info(
+                )->execute();
+
+            if (false === $del) {
```
tomasnorre: Did I overwrite you log change @bmack ?

bmack: looks ok

bmack: oh yes, you did. instead of `getLogger()->log` use `logger->info(`

PR #430, opened 2019-08-22
[FEATURE] Use new ContextMenu API instead of ClickMenu logic

bmack:
The new ItemProvider for ContextMenu built in TYPO3 v7/v8
is now used, and all old ClickMenu logic is removed.

Fixes: 429


Classes/ContextMenu/ItemProvider.php
```diff
@@ -0,0 +1,98 @@
+<?php
+declare(strict_types = 1);
+namespace AOE\Crawler\ContextMenu;
+
+/***************************************************************
+ *  Copyright notice
+ *
+ *  (c) 2017 AOE GmbH <dev@aoe.com>
```
tomasnorre: 2019

tomasnorre: I'll run over the license later.

PR #437, opened 2019-08-22
[BUGFIX] Make sure that Process is added and mapped correctly (#413)

### LINKED ISSUE #413: [BUG] Crawler limitation "Count in a run" does not work
I did a "Crawler Run" with a 5 page limit (Count in a run). Upon parsing the first page, the crawler generates a new crawler process and assigns it another 5 pages, overwriting the last 4 unhandled pages in the queue as well. Thus after each page-fetch, a new process gets spawned and the page fetch-limit resets every time. In the end, every page gets fetched instead of stopping after grabbing the first 5.

The issue seems to be, that while the query to find the process completes with results, all results are basically clean processes with no values set:
```
array(10) {
    ["row"] => NULL
    ["processId"] => string(0) ""
    ["active"]  => bool(false)
    ["ttl"] => int(0)
    ["assignedItemsCount"] => int(0)
    ["deleted"] => bool(false)
    ["systemProcessId"] => string(0) ""
    ["queueRepository"] => NULL
    ["uid"] =>int(0)
    ["pid"] =>int(0)
  }
```

Checking the raw query results (`$query->execute(true)`) does show the missing values. they are just not being assigned to the model correctly...

_Originally posted by @Ceremony64 in https://github.com/AOEpeople/crawler/issues/408#issuecomment-523381985_

tomasnorre:
Resolves #413

### DIFF
--- a/Classes/Backend/View/ProcessListView.php
+++ b/Classes/Backend/View/ProcessListView.php
@@ -94,6 +94,11 @@ class ProcessListView
     protected $assignedUnprocessedItemCount;
 
     /**
+     * @var ProcessCollection
+     */
+    protected $processCollection;
+
+    /**
      * @return int
      */
     public function getAssignedUnprocessedItemCount()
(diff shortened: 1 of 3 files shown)


bmack:
👍

PR #440, opened 2019-08-27
Rename the CLI user

### LINKED ISSUE #439: Rename the CLI user
The user which is needed to run the CLI jobs has been renamed in TYPO3 8.7:  
https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/8.6/Feature-79240-SingleCliUserForCliCommands.html

Tasks:
- All occurences in EXT:crawler need to be changed.
- Ensure compatibility with older versions (or add a conflict with TYPO3 v7 and below).
- Remove admin permission check / ignore the fact that the new `_cli_` user has admin permissions out of the box.

stucki:
Closes #439

### DIFF
--- a/Classes/Backend/BackendModule.php
+++ b/Classes/Backend/BackendModule.php
@@ -1016,7 +1016,7 @@ class BackendModule extends AbstractFunctionModule
     }
 
     /**
-     * Indicate that the required be_user "_cli_crawler" is
+     * Indicate that the required be_user "_cli_" is
      * global available in the system.
      *
      * @return boolean
@@ -1024,7 +1024,7 @@ class BackendModule extends AbstractFunctionModule
     protected function isCrawlerUserAvailable()
     {
         $isAvailable = false;
-        $userArray = BackendUtility::getRecordsByField('be_users', 'username', '_cli_crawler');
+        $userArray = BackendUtility::getRecordsByField('be_users', 'username', '_cli_');
 
         if (is_array($userArray)) {
             $isAvailable = true;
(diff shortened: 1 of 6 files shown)


tomasnorre:
Thanks for you PR @stucki , as we are still having support for TYPO3 V7 this can not be merged yet.

stucki:
@tomasnorre Thanks for your feedback. The thing is that this currently works as long as the `_cli_crawler` user is created manually, but it will be deleted during updates from 7.6 to 8.7 (see https://github.com/TYPO3/TYPO3.CMS/blob/TYPO3_8-7/typo3/sysext/install/Classes/Updates/CommandLineBackendUserRemovalUpdate.php). So it is ok to do nothing, but this should be avoided since TYPO3 8.7 expects to have only one `_cli_` user.

These are our options:
- We can put the PR on hold and merge it as soon as 7.6 support is dropped
- otherwise, split the master branch
- or add conditions into the code

What do you prefer?

tomasnorre:
I prefer having it on hold. 
I would expect dropped support end of Year.

My second best option would be to branch, and release from there.

What is your opinion on this ?

tomasnorre:
Could an option be having a migration wizards, that provides the "old" user again, or will this result in an "infinite" loop with to wizards, one deleting and one adding the user.

tomasnorre:
Hi @stucki,

If you could rebase your changes against typo3v9 branch, we can have it merged already and have this PR closed.

Thanks for you help.
Happy New Year.

tomasnorre:
Hi @stucki 

We are about to release the 9.0.0 version, and switched the branches around. This closed this PR automatically. 

Can you please recreate it with reference to `TYPO3_7` instead.
Thanks

stucki:
@tomasnorre Thanks, just checked. The issue is that the `TYPO3_7` branch should support both TYPO3 v7 and v8 but the user still needs to be added in v7. However, as it turns out, the change was merged (in a different way) into the latest master branch. So IMO there's nothing left to do and this PR can be closed.

tomasnorre:
@stucki 
FYI: The master branch will not have support for TYPO3 8 and TYPO3 7. 
Currently for TYPO3 9 and 10.

stucki:
@tomasnorre I know. The issue remains unfixed for TYPO3 v8, but that's ok for me. (I used a local patch which can be removed when going up to the latest release...)

Issue #446, opened 2019-09-02
[BUG] cliPath is only working with composer-installs


tomasnorre:
Currently the `getCrawlerCliPath()` is only working with composer, not all installation are composer based, so we need a way to resolve this not depending on composer only. 

https://github.com/AOEpeople/crawler/blob/typo3v9/Classes/Service/ProcessService.php#L223

tomasnorre:
This is done for dev-master in https://github.com/AOEpeople/crawler/pull/449

tomasnorre:
Also resolved in https://github.com/AOEpeople/crawler/tree/typo3v9

PR #449, opened 2019-09-02
[WIP][BUGFIX] Resolving typo3cms path if composer not used (#446)

tomasnorre:


tstahn:
Just a thought/general question after I've checked on these changes:

Is there any reason why we are not dropping the typo3-console usage/dependency completely in favor of the TYPO3 Cores `typo3/sysext/core/bin/typo3` binary?

As I see it this would have several benefits:
- no composer setup required
- no typo3-console required
- native TYPO3 support (no 3rd party dependencies)
- low threshold for (new )users as there is no need to learn about typo3-console first
- no need for path mapping/finding inside the crawler extension
- no need for checking whether or not crawler is used in composer environment

Last but not least:
- less overhead
- less potential for things to break

tomasnorre:
@tstahn 

I agree on the points here. 
But would not like to switch in the 6.x versions. 

I would prefer only doing this switch in the V9. 
Don't want to introduce to0 many new things in a version, that will soon be obsolete.

But I'll consider it, because you're totally right on your point, and if someone decides to use the typo3cms (typo3-console) it still working, but not a hard dependency anymore..

But we cannot do that because of symfony/console before we drop support for TYPO3V7, that's also one of my reason to only do it for the TYPO3V9 version.

Classes/Service/ProcessService.php
```diff
@@ -225,20 +226,26 @@ public function startProcess()
     public function getCrawlerCliPath()
     {
         $composerRootDir = getenv('TYPO3_PATH_COMPOSER_ROOT') . '/';
-        $jsonDecoded = json_decode(file_get_contents($composerRootDir . 'composer.json'), true);
+        $composerFile = $composerRootDir . 'composer.json';
+        $phpPath = $this->crawlerController->extensionSettings['phpPath'] . ' ';
+        $cliPart = 'typo3cms crawler:crawlqueue';
+
+        if (file_exists($composerFile)) {
+            $jsonDecoded = json_decode(file_get_contents($composerRootDir . 'composer.json'), true);
```
tstahn: You could have re-used `$composerFile` here.

tomasnorre: Sure

PR #451, opened 2019-09-04
[TASK] Streamline Configuration fetching and URL building

bmack:
Fetching crawler configuration from the database and the according URL building process is a bit cleaner now:

- Fetching configuration from the rootline is now in a separate method. Easier for mocking in tests
- The method parseParams() has been replaced by GeneralUtility:: explodeUrl2Array()
- URL generation for variants of a page ID has been streamlined and building more easily
- Minor foreach / array checks are simplified

This is a precursor to streamline further URL compilation changes

### DIFF
--- a/Classes/Controller/CrawlerController.php
+++ b/Classes/Controller/CrawlerController.php
@@ -180,7 +180,7 @@ class CrawlerController implements LoggerAwareInterface
     /**
      * @var int
      */
-    protected $maximumUrlsToCompile;
+    protected $maximumUrlsToCompile = 10000;
 
     /**
      * Method to set the accessMode can be gui, cli or cli_im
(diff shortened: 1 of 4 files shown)


Classes/Controller/CrawlerController.php
```diff
@@ -2561,4 +2503,34 @@ protected function getConfigurationHash(array $configuration)
         unset($configuration['URLs']);
         return md5(serialize($configuration));
     }
+
+    /**
+     * Traverses up the rootline of a page and fetches all crawler records
+     * @param int $pageId
+     * @return array
+     */
+    protected function getCrawlerConfigurationRecordsFromRootLine(int $pageId): array
```
tomasnorre: I would prefer is this is done from the Repository.
https://github.com/AOEpeople/crawler/blob/typo3v9/Classes/Domain/Repository/ConfigurationRepository.php

bmack: Done

Issue #443, opened 2019-09-02
[FEATURE] Use Site Handling instead of sys_domain


tomasnorre:
With this feature we will remove the dependency for sys_domain, and use the Site Handling coming with TYPO3 V9 instead.

tomasnorre:
Resolved in #452

PR #452, opened 2019-09-04
[!!!][FEATURE] Automatically calculate cHash and Base URL via Site Handling

bmack:
When using Site Handling in TYPO3 v9, pages with a site component now receive a baseUrl from the current site where the page is located on.

Using Site Handling's PageRouter functionality, the cHash is then automatically added when needed.

In addition to that, when having no site configuration (which is only the case when dealing with RealURL v9 forks) the cHash is then automatically calculated.

For this case, the option "tx_crawler_configuration.cHash" is removed without substitution, to follow TYPO3 Core's pattern to have TYPO3 Core deal with cHash generation then.


tomasnorre:
This resolves #443

PR #438, opened 2019-08-23
[FEATURE] Setting up Domain/Model+Repository

tomasnorre:

PR #454, opened 2019-09-04
[!!!][FEATURE] Do not rely on sys_domain anymore but use Site Handling

bmack:
By using site handling by default, it is still possible to override the base URL via the "base_url" value in a crawler configuration, but if left empty, the proper site handling option is selected automatically.

The database field tx_crawler_configuration.sys_domain_base_url and all functionality is therefore removed.

All fixtures now also contain proper Pages with slugs and language information, as defined in TYPO3 v9.

Fixes: 443


Classes/Controller/CrawlerController.php
```diff
@@ -441,56 +441,62 @@ public function urlListFromUrlArray(
...
+            // Scheduled time:
+            $schTime = $scheduledTime + round(count($duplicateTrack) * (60 / $reqMinute));
+            $schTime = floor($schTime / 60) * 60;
 
-                        $urlList .= '<br />';
+            if (isset($duplicateTrack[$uKey])) {
+                //if the url key is registered just display it and do not resubmit is
+                $urlLog[] = '<em><span class="text-muted">' . htmlspecialchars($url) . '</span></em>';
+            } else {
+                $urlList = '[' . date('d.m.y H:i', $schTime) . '] ' . htmlspecialchars($url);
```
tomasnorre: Think I would like to have this configurable, perhaps using the default TYPO3 

`[SYS][ddmmyy] [SYS][hhmm]` and the `[SYS][USdateFormat]` from LocalConfiguration, so that one decides for itself how the date looks like.

Is not a must have, but just an idea.

Classes/Controller/CrawlerController.php
```diff
@@ -612,21 +613,14 @@ public function getUrlsForPageId($pageId)
                         $subCfg = [
                             'procInstrFilter' => $configurationRecord['processing_instruction_filter'],
                             'procInstrParams.' => $TSparserObject->setup,
-                            'baseUrl' => $this->getBaseUrlForConfigurationRecord(
-                                $configurationRecord['base_url'],
-                                (int)$configurationRecord['sys_domain_base_url'],
-                                (bool)($configurationRecord['force_ssl'] > 0)
-                            ),
+                            'baseUrl' => $configurationRecord['base_url'],
```
tomasnorre: This would conflict with my change in: 
https://github.com/AOEpeople/crawler/pull/438

I only need a few changed before I can merged, after review from @tstahn

Classes/Controller/CrawlerController.php
```diff
@@ -612,21 +613,14 @@ public function getUrlsForPageId($pageId)
...
+                            'force_ssl' => (int)$configurationRecord['force_ssl'],
```
tomasnorre: Didn't you refactor this force_ssl already, Think I saw something, would need to look back into that.. 

Think you change the behaviour here lately.

bmack: Yeah, I broke it, now I fixed it again :)

Classes/Controller/CrawlerController.php
bmack: yup. left it as before, but we can change that ofc.

bmack: done

PR #455, opened 2019-09-04
[FEATURE] Migrate FE hooks to PSR-15 middleware

bmack:
Used hooks in TSFE are deprecated, and PSR-15 middlewares and PSR-7 Request/Response data allows
for sending and evaluating crawler instructions in one go.

ToDo:
- We need to check "_CLI_crawler" and the ominuous Backend user Auth which is probably unneeded.
- We can probably just initialize BE_USER in the same middleware.


bmack:
This one isn't tested yet, so let's wait a bit please.

Classes/Domain/Repository/QueueRepository.php
```diff
@@ -511,4 +511,17 @@ public function getAvailableSets(): array
...
         return $rows;
     }
+
+    public function findByQueueId(string $queueId): ?array
+    {
+        $queueRec = $this->queryBuilder
+            ->select('*')
+            ->from($this->tableName)
+            ->where(
+                $this->queryBuilder->expr()->eq('qid', $this->queryBuilder->createNamedParameter($queueId))
```
tomasnorre: shouldn't we add `\PDO::PARAM_STR` I'm not 100% when it's recommend and why not.

Classes/Middleware/CrawlerInitialization.php
```diff
@@ -0,0 +1,143 @@
...
+     */
+    protected $queueRepository;
+
+    /**
+     * @param Context|null $context
+     */
+    public function __construct(Context $context = null)
+    {
+        $this->context = $context ?? GeneralUtility::makeInstance(Context::class);
+        $this->queueRepository = GeneralUtility::makeInstance(QueueRepository::class);
```
tomasnorre: Why use the GeneralUtility and not the ObjectManager what is the win of this way, and when is GeneralUtility preferred over ObjectManager and vice versa.

Classes/Middleware/CrawlerInitialization.php
```diff
@@ -0,0 +1,143 @@
...
+        ];
+
+        $this->runPollSuccessHooks();
+
+        // Output log data for crawler (serialized content):
+        $content = serialize($GLOBALS['TSFE']->applicationData['tx_crawler']);
+        return new Response($content, 200);
+    }
+
+    protected function isRequestHashMatchingQueueRecord(?array $queueRec, string $hash): bool
```
tomasnorre: Is the `?` here correct, don't recall seeing that notation before.

bmack: yup. meaning array or null. PHP 7.1 feature. We return "null" from the QueueRepository that's why it is checked here.

Classes/Middleware/CrawlerInitialization.php
bmack: If you have inject* methods you need to use ObjectManager. But it adheres a dependency on Extbase, so GeneralUtility is sufficient in most cases.

In some cases ObjectManager needs TypoScript configuration to fetch the right objects (in Extbase plugins mostly), which you see this more in third-party-extensions based on Extbase,

Classes/Domain/Repository/QueueRepository.php
bmack: `\PDO::PARAM_STR` is the default argument.

Classes/Middleware/CrawlerInitialization.php
```diff
@@ -57,7 +58,7 @@ class CrawlerInitialization implements MiddlewareInterface
     public function __construct(Context $context = null)
     {
         $this->context = $context ?? GeneralUtility::makeInstance(Context::class);
```
tomasnorre: This is also GeneralUtility

Classes/Domain/Repository/QueueRepository.php
tomasnorre: Didn't know that.

bmack:
Works fine, all tested now.

PR #456, opened 2019-09-04
[FEATURE] Switch to symfony/console and typo3 binary (#450)

### LINKED ISSUE #450: [FEATURE] Switch to the Symfony Command
Instead of using the `Helhum\Typo3Console\Mvc\Controller\CommandController` we should use the Symfony Command instead.

See: https://typo3worx.eu/2019/01/symfony-console-in-typo3/

- [x] Switch to Symfony Command
- [x] Use the typo3 binary instead of typo3cms
- [x] Adjust the ProcessService->getCrawlerCliPath() to new binary

tomasnorre:
Resolves #450


Classes/Command/FlushQueueCommand.php
```diff
@@ -0,0 +1,112 @@
...
+        );
+        $this->addOption(
+            'mode',
+            'm',
+            InputOption::VALUE_OPTIONAL,
+            'Output mode',
+            'finished'
+		);
+
+        $this->addOption(
```
bmack: above it's "startpage" / "s", here it is "p" - maybe we should go with "pid"/"p" or "page"/"p" as option name.

tomasnorre: Which do you want to correct the `startpage` or this one ?

PR #457, opened 2019-09-04
[!!!][FEATURE] Separate Execution of URLs into strategy classes

bmack:
The following public methods have beem removed:
- CrawlerController->readUrl_exec()
- CrawlerController->requestUrl()

The following option is removed:
- follow30x (as this is handled by guzzle now)

Instead of having all logic within CrawlerController, the logic to execute a URL from a QueueItem is moved into the QueueExecutor. A queue record is handed over which then validates against Extension settings which Strategy to choose (Guzzle, Callback for external files, SubProcess via custom PHP script).

As the Guzzle RequestFactory from TYPO3 Core is used, all default settings (e.g. follow redirects) are applied directly, so EXT:crawler does not have to have a custom option anymore.

Fixes: 330


Classes/CrawlStrategy/CallbackExecutionStrategy.php
```diff
@@ -0,0 +1,41 @@
+<?php
+declare(strict_types=1);
+namespace AOE\Crawler\CrawlStrategy;
+
+/*
```
tomasnorre: Could you please use the aoe-license text.

```
/***************************************************************
 *  Copyright notice
 *
 *  (c) 2019 AOE GmbH <dev@aoe.com>
 *
 *  All rights reserved
 *
 *  This script is part of the TYPO3 project. The TYPO3 project is
 *  free software; you can redistribute it and/or modify
 *  it under the terms of the GNU General Public License as published by
 *  the Free Software Foundation; either version 3 of the License, or
 *  (at your option) any later version.
 *
 *  The GNU General Public License can be found at
 *  http://www.gnu.org/copyleft/gpl.html.
 *
 *  This script is distributed in the hope that it will be useful,
 *  but WITHOUT ANY WARRANTY; without even the implied warranty of
 *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 *  GNU General Public License for more details.
 *
 *  This copyright notice MUST APPEAR in all copies of the script!
 ***************************************************************/
```

Classes/CrawlStrategy/GuzzleExecutionStrategy.php
```diff
@@ -0,0 +1,81 @@
+<?php
+declare(strict_types=1);
+namespace AOE\Crawler\CrawlStrategy;
+
+/*
```
tomasnorre: Dito

Classes/CrawlStrategy/SubProcessExecutionStrategy.php
```diff
@@ -0,0 +1,157 @@
...
+
+        if (!is_array($parsedUrl)) {
+            return [];
+        }
+
+        $requestHeaders = $this->buildRequestHeaders($parsedUrl, $crawlerId);
+
+        $cmd = escapeshellcmd($this->extensionSettings['phpPath']);
+        $cmd .= ' ';
+        $cmd .= escapeshellarg(ExtensionManagementUtility::extPath('crawler') . 'cli/bootstrap.php');
```
tomasnorre: I think the `cli/bootstrap.php` is obsolete and could be bypassed ?

tomasnorre: I was expecting this to happen with the Guzzle

Classes/CrawlStrategy/SubProcessExecutionStrategy.php
```diff
@@ -0,0 +1,157 @@
...
+
+    /**
+     * Executes a shell command and returns the outputted result.
+     *
+     * @param string $command Shell command to be executed
+     * @return string Outputted result of the command execution
+     */
+    protected function executeShellCommand($command)
+    {
+        return shell_exec($command);
```
tomasnorre: Why not use the `CommandUtility::exec` ? and use that directly where needed.

Classes/QueueExecutor.php
```diff
@@ -0,0 +1,102 @@
...
+            $result = GeneralUtility::makeInstance(CallbackExecutionStrategy::class)
+                ->fetchByCallback($className, $parameters, $crawlerController);
+            $result = ['content' => serialize($result)];
+        } else {
+            // Regular FE request
+            $crawlerId = $this->generateCrawlerIdFromQueueItem($queueItem);
+
+            // Get result:
+            $url = new Uri($parameters['url']);
+            $result = $this->selectedStrategy->fetchUrlContents($url, $crawlerId);
```
tomasnorre: Ahh . That's why we still need it with `cli/bootstrap.php`..

Classes/QueueExecutor.php
```diff
@@ -0,0 +1,102 @@
...
+            EventDispatcher::getInstance()->post(
+                'urlCrawled',
+                $queueItem['set_id'],
+                ['url' => $parameters['url'], 'result' => $result]
+            );
+        }
+        return $result;
+    }
+
+    protected function generateCrawlerIdFromQueueItem(array $queueItem): string
```
tomasnorre: I like that you are using the encryptionkey, but doesn't we have this somewhere already with generateCrawlerId

Tests/Unit/QueueExecutorTest.php
```diff
@@ -0,0 +1,69 @@
...
+    public function validateTypo3InternalGuzzleExecutionIsSelected()
+    {
+        $configuration = [
+            'makeDirectRequests' => 0,
+            'frontendBasePath' => '/',
+        ];
+        $GLOBALS['TYPO3_CONF_VARS']['EXTENSIONS']['crawler'] = $configuration;
+        $subject = $this->getAccessibleMock(QueueExecutor::class, ['dummy']);
+        $res = $subject->_get('selectedStrategy');
+        self::assertEquals(get_class($res), GuzzleExecutionStrategy::class);
```
tomasnorre: I always to `$this->assertEquals()` any benefit doing `self::assertEquals()`

ext_conf_template.txt
```diff
@@ -5,9 +5,6 @@
 # cat=Settings; type=string; label=Frontend website base path: Base path of the website frontend (e.g. if you call http://mydomain.com/cms/index.php in the browser the base path is "/cms/"). Leave empty to use the value of config.absRefPrefix instead.
 frontendBasePath=/
 
-# cat=Settings; type=boolean; label= Follow 301 and 302: If checked the crawler will follow the URLs (Location) given in a HTTP 301 and 302 response.
-follow30x=
-
 # cat=Settings; type=boolean; label= Crawl hidden pages: Crawl hidden pages (By default they won't be crawled)
 crawlHiddenPages=0
```
tomasnorre: I think the `makeDirectRequest` in here can be removed as well.

bmack: nope. we need to keep this to distinct between our own subprocess and guzzle

Tests/Unit/QueueExecutorTest.php
bmack: assertEquals is defined as static in phpunit, that's why I used this.

Classes/QueueExecutor.php
bmack: yes

Classes/CrawlStrategy/SubProcessExecutionStrategy.php
bmack: wanted to do that with the next patch

Classes/CrawlStrategy/CallbackExecutionStrategy.php
bmack: yeah, can do that. However, I'd rather stay with GPL2+ but apparantly this repo is GPL3 only.

Classes/CrawlStrategy/SubProcessExecutionStrategy.php
tomasnorre: Should we add a Task for it, so we don't forget?

bmack: no, that's https://github.com/AOEpeople/crawler/pull/432 - which I'd like to rebase once this patch is in.

PR #462, opened 2019-09-05
[TASK] Clean up code in CrawlerController

bmack:
- Use proper (int) type-casting instead of intval() = faster
- Use $this->tableName instead of 'tx_crawler_queue'
- Use $this->backendUser instead of $GLOBALS[BE_USER]
- Remove unused HTML styling code
- Move CLI_checkIfProcessIsActive into ProcessRepository

### DIFF
--- a/Classes/Controller/CrawlerController.php
+++ b/Classes/Controller/CrawlerController.php
@@ -47,6 +47,7 @@ use TYPO3\CMS\Core\Imaging\Icon;
 use TYPO3\CMS\Core\Imaging\IconFactory;
 use TYPO3\CMS\Core\Routing\SiteMatcher;
 use TYPO3\CMS\Core\Site\Entity\Site;
+use TYPO3\CMS\Core\Type\Bitmask\Permission;
 use TYPO3\CMS\Core\TypoScript\Parser\TypoScriptParser;
 use TYPO3\CMS\Core\Utility\DebugUtility;
 use TYPO3\CMS\Core\Utility\GeneralUtility;
@@ -407,7 +408,7 @@ class CrawlerController implements LoggerAwareInterface
             ->count('*')
             ->from($this->tableName)
             ->where(
-                $queryBuilder->expr()->eq('page_id', intval($uid)),
+                $queryBuilder->expr()->eq('page_id', (int)$uid),
                 $queryBuilder->expr()->eq('configuration_hash', $queryBuilder->createNamedParameter($configurationHash)),
                 $queryBuilder->expr()->eq('exec_time', 0)
             )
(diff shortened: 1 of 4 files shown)


Classes/Domain/Repository/ProcessRepository.php
```diff
@@ -296,4 +296,20 @@ public function deleteProcessesMarkedAsDeleted()
             )
             ->execute();
     }
+
+    public function isProcessActive(int $processId): bool
+    {
+        $queryBuilder = GeneralUtility::makeInstance(ConnectionPool::class)->getQueryBuilderForTable($this->tableName);
+        $isActive = $queryBuilder
+            ->select('active')
+            ->from('tx_crawler_process')
```
tomasnorre: `$this->tableName` ?  it's already used two lines above.

Classes/Domain/Repository/ProcessRepository.php
```diff
@@ -296,4 +296,20 @@ public function deleteProcessesMarkedAsDeleted()
...
+            ->where(
+                $queryBuilder->expr()->eq('process_id', $queryBuilder->createNamedParameter($processId, \PDO::PARAM_INT))
```
tomasnorre: The `process_id` is not an integer.. 
https://github.com/AOEpeople/crawler/pull/462/files#diff-481209dc0ee6be420867b4ac30992471R1980

Classes/Controller/CrawlerController.php
```diff
@@ -1360,12 +1361,12 @@ public function readUrlFromArray($field_array)
     {
             // Set exec_time to lock record:
         $field_array['exec_time'] = $this->getCurrentTime();
-        $connectionForCrawlerQueue = GeneralUtility::makeInstance(ConnectionPool::class)->getConnectionForTable('tx_crawler_queue');
+        $connectionForCrawlerQueue = GeneralUtility::makeInstance(ConnectionPool::class)->getConnectionForTable($this->tableName);
```
tomasnorre: I know I started this code, but perhaps it should be `$this->tableCrawleQueue` instead, as we might end up with more tableNames later. 

But for now it isn't a problem and can be refactored later.

Classes/Domain/Repository/ProcessRepository.php
bmack: correct. will fix

Classes/Domain/Repository/ProcessRepository.php
tomasnorre: Ups.. Then it's on me. But shouldn't work.

bmack: fixed.

PR #463, opened 2019-09-11
[WIP][BUGFIX] Add process in Backend

tomasnorre:


Classes/Utility/PhpBinaryUtility.php
```diff
@@ -0,0 +1,42 @@
+<?php
```
tomasnorre: I don't really like this approach, but didn't have better ideas in my mind.
Please point me to some direction how this can be solved.

Classes/Service/ProcessService.php
```diff
@@ -226,10 +227,11 @@ public function startProcess()
      */
     public function getCrawlerCliPath(): string
     {
-        $phpPath = $this->crawlerController->extensionSettings['phpPath'] . ' ';
-        $cliPart = 'typo3 crawler:processQueue';
+        $phpPath = PhpBinaryUtility::getPhpBinary($this->crawlerController->extensionSettings);
```
tomasnorre: https://github.com/AOEpeople/crawler/pull/463/files#diff-bf56d9779b9916e69b145ec4d789c571R209 isn't working. 

The process is not added, expecting it to be due to problems with the `commandUtility::exec` and the subProcesses.. But could be wrong.

Classes/Service/ProcessService.php
```diff
@@ -200,9 +201,9 @@ public function startProcess()
 
         // Check whether OS is Windows
         if (Environment::isWindows()) {
-            $completePath = CommandUtility::escapeShellArgument('start ' . $this->getCrawlerCliPath());
+            $completePath = 'start ' . $this->getCrawlerCliPath();
```
tomasnorre: The CommandUtility::escapeShellArgument breaks the funktionality, Or I'm doing something wrong. 

It ends up with extra `'` and cannot be executed. At least not on Unix, didn't test on windows.

CDRO: I think this is because of the `getCrawlerCliPath` that is already escaped, therefore double-escaping the content?

Classes/Utility/PhpBinaryUtility.php
CDRO: I'd not pass the extension settings on each call, I thing it's absolutely ok to just call `getPhpBinary` and solely rely on Extension configuration (which can be fetched as described [here](https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/ExtensionArchitecture/ConfigurationOptions/Index.html)). Or is there any need to pass it on each call?

tomasnorre: I need both `phpPath` and `phpBinary` to ensure I can provide the correct information in case of phpPath is set.

Classes/Service/ProcessService.php
tomasnorre: I think you're right.

Classes/Utility/PhpBinaryUtility.php
tomasnorre: You're right, I don't need to pass it, it can be fetch inside the function directly.

PR #466, opened 2019-09-24
Refactor check for php executable

stefa50:
CommandUtility delivers possibility to check for an executable command, so let's use this one. By the way the current code fails due to wrong use of CommandUtility::escapeShellArgument for escaping the command not the arguments.


tomasnorre:
@stefa50 Thanks for you PR, We already have a work in progress for this.
https://github.com/AOEpeople/crawler/pull/463

Can you please check and give your comments on this?
Would be helpful in order to have it merged.

Thanks

tomasnorre:
This should now be fixed in:
https://github.com/AOEpeople/crawler/pull/463

Please reopen if this is not the case.

Issue #465, opened 2019-09-11
[BUG] Status isn't updated in Crawler Log


tomasnorre:
Steps to reproduce:

1. Go to the Crawler Log
2. Press the "process wheal"
3. Status column isn't updated, at least not on error.

FYI: You can verify the status, but clicking at the Queue Id and see the result data. 

![Screenshot 2019-09-11 at 21 36 51](https://user-images.githubusercontent.com/1212481/64728901-58d8af00-d4dc-11e9-9d95-9b9860b25d92.png)

PR #480, opened 2019-11-27
[BUGFIX] Update error message on error with GuzzleRequests

tomasnorre:


tomasnorre:
Resolves: #465

composer.json
```diff
@@ -24,7 +24,8 @@
         "typo3/cms-frontend": "*",
         "typo3/cms-info": "*",
         "typo3/cms-extensionmanager": "*",
-        "symfony/console": "~4.3"
+        "symfony/console": "~4.3",
+        "guzzlehttp/guzzle": "6.4.1"
```
bmack: I think we should up for using "^6.4.1" or something the like. Same with symfony/console (^4.3)

Issue #479, opened 2019-11-27
[FEATURE] add _RECURSIVE option to the subpartParams Configuration of the lookup _TABLE instructions


tomasnorre:
## Bug Report

Please check if this also applies for the typo3v9 branch.
https://github.com/AOEpeople/crawler/pull/477

tomasnorre:
@vzz3 Do you have the possibility to check if this also applies to the typo3v9 branch ?

vzz3:
I can probably implement that for the typo3v9 branch as well. However, I don't know how I should test it. Does the head of the typo3v9 branch already run under TYPO3 9 LTE or should I test with TYPO3 8 LTE?

tomasnorre:
The typo3v9 branch is already running under TYPO3 9 LTS, so if you have a 9 setup somewhere, it should be fairly easy to test it.

Thanks for helping out. 
There is even a ddev setup for the crawler, so it should be fairly easy to set up for testing.
https://github.com/AOEpeople/crawler-devbox/tree/feature/ddev

tomasnorre:
This is merged, will close the issue. 
Thanks @vzz3

PR #483, opened 2019-11-28
[FEATURE] add _RECURSIVE option to the subpartParams Configuration of the lookup _TABLE instructions (#479)

vzz3:
Applying my patch [eef0c94] to the typo3v9 branch was easy. Fixing some bugs that prevented me from testing the new table based parameter expansion and applying 2 change sets (18.07.19 [98605ee], 22.07.19 [e17ec22]) that were required for automated testing was the harder part.

Pleas have a look especially  at the changes of the methods getConfigurationsForBranch() and getDuplicateRowsIfExist () of the class CrawlerController in commit [91e88fc]. I changed them in in a way that the fulfil the test cases from the TYPO3 v8 (current master) branch , but I'm not sure if this is intended.

### DIFF
--- a/Classes/Command/BuildQueueCommand.php
+++ b/Classes/Command/BuildQueueCommand.php
@@ -128,12 +128,7 @@ re-indexing or static publishing from command line.' . chr(10) . chr(10) .
             $crawlerController->registerQueueEntriesInternallyOnly = true;
         }
 
-        #if ($this->request instanceof \TYPO3\CMS\Extbase\Mvc\Cli\Request) {
-            $pageId = MathUtility::forceIntegerInRange($input->getOption('page'), 0);
-        #} else {
-            // Crawler is called over Backend
-        #    $pageId = 1;
-        #}
+        $pageId = MathUtility::forceIntegerInRange($input->getOption('page'), 0);
 
         $configurationKeys = $this->getConfigurationKeys($input->getOption('conf'));
 
--- a/Classes/Command/BuildQueueCommand.php
+++ b/Classes/Command/BuildQueueCommand.php
@@ -128,12 +128,12 @@ re-indexing or static publishing from command line.' . chr(10) . chr(10) .
             $crawlerController->registerQueueEntriesInternallyOnly = true;
         }
 
-        if ($this->request instanceof \TYPO3\CMS\Extbase\Mvc\Cli\Request) {
-            $pageId = MathUtility::forceIntegerInRange($input->getOption('startpage'), 0);
-        } else {
+        #if ($this->request instanceof \TYPO3\CMS\Extbase\Mvc\Cli\Request) {
+            $pageId = MathUtility::forceIntegerInRange($input->getOption('page'), 0);
+        #} else {
             // Crawler is called over Backend
-            $pageId = 1;
-        }
+        #    $pageId = 1;
+        #}
 
         $configurationKeys = $this->getConfigurationKeys($input->getOption('conf'));
 
--- a/Classes/Controller/CrawlerController.php
+++ b/Classes/Controller/CrawlerController.php
@@ -141,9 +141,9 @@ class CrawlerController implements LoggerAwareInterface
     protected $accessMode;
 
     /**
-     * @var BackendUserAuthentication
+     * @var BackendUserAuthentication|null
      */
-    private $backendUser;
+    private $backendUser = null;
 
     /**
      * @var integer
@@ -278,7 +278,6 @@ class CrawlerController implements LoggerAwareInterface
         $this->queueExecutor = $objectManager->get(QueueExecutor::class);
         $this->iconFactory = GeneralUtility::makeInstance(IconFactory::class);
 
-        $this->backendUser = $GLOBALS['BE_USER'];
         $this->processFilename = Environment::getVarPath() . '/locks/tx_crawler.proc';
 
         /** @var ExtensionConfigurationProvider $configurationProvider */
@@ -293,6 +292,16 @@ class CrawlerController implements LoggerAwareInterface
 
         $this->extensionSettings['processLimit'] = MathUtility::forceIntegerInRange($this->extensionSettings['processLimit'], 1, 99, 1);
         $this->maximumUrlsToCompile = MathUtility::forceIntegerInRange($this->extensionSettings['maxCompileUrls'], 1, 1000000000, 10000);
+    }
+    
+    /**
+     * @return BackendUserAuthentication
+     */
+    private function getBackendUser() {
+        if($this->backendUser === null) {
+            $this->backendUser = $GLOBALS['BE_USER'];
+        }
+        return $this->backendUser;
     }
 
     /**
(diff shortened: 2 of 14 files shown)


tomasnorre:
Thanks for your effort @vzz3.
I'll look into this beginning of next week I hope.

Classes/Command/BuildQueueCommand.php
```diff
@@ -128,12 +128,12 @@ protected function execute(InputInterface $input, OutputInterface $output)
             $crawlerController->registerQueueEntriesInternallyOnly = true;
         }
 
-        if ($this->request instanceof \TYPO3\CMS\Extbase\Mvc\Cli\Request) {
-            $pageId = MathUtility::forceIntegerInRange($input->getOption('startpage'), 0);
-        } else {
+        #if ($this->request instanceof \TYPO3\CMS\Extbase\Mvc\Cli\Request) {
```
tomasnorre: That's not how to comment code in PHP :) `//`
But please don't just comment out code if it's not needed anymore remove it.

Classes/Controller/CrawlerController.php
```diff
@@ -657,7 +666,7 @@ public function getUrlsForPageId($pageId)
      * @param $depth
      * @return array
      *
-     * TODO: Write Functional Tests
+     * TODO: Write Functional Tests => done?!?!
```
tomasnorre: If I remember correctly, it isn't done yet..

Classes/Controller/CrawlerController.php
```diff
@@ -677,18 +686,21 @@ public function getConfigurationsForBranch(int $rootid, $depth)
         }
         /* @var PageTreeView $tree */
         $tree = GeneralUtility::makeInstance(PageTreeView::class);
-        $perms_clause = $this->backendUser->getPagePermsClause(Permission::PAGE_SHOW);
-        $tree->init('AND ' . $perms_clause);
+        $perms_clause = $this->getBackendUser()->getPagePermsClause(Permission::PAGE_SHOW);
+        $tree->init( empty($perms_clause) ? '' : ('AND ' . $perms_clause) );
```
tomasnorre: 👍

Classes/Controller/CrawlerController.php
```diff
@@ -677,18 +686,21 @@ public function getConfigurationsForBranch(int $rootid, $depth)
...
         foreach ($tree->tree as $node) {
             $pids[] = $node['row']['uid'];
         }
 
         $queryBuilder = $this->getQueryBuilder('tx_crawler_configuration');
 
-        $queryBuilder->getRestrictions()
-            ->removeAll()
-            ->add(GeneralUtility::makeInstance(DeletedRestriction::class));
+        // The pre TYPO3 v9 version use the method BackendUtility::BEenableFields('tx_crawler_configuration')
```
tomasnorre: If not needed, please remove

tomasnorre:
Added a few comments, but will look into it next week.

vzz3:
Please tell me when you're done with the review, then I like to fix all changes at once.

Classes/Controller/CrawlerController.php
```diff
@@ -810,25 +823,36 @@ public function expandParameters($paramArray, $pid)
                             if ($fieldName === 'uid' || $GLOBALS['TCA'][$subpartParams['_TABLE']]['columns'][$fieldName]) {
                                 $queryBuilder = $this->getQueryBuilder($subpartParams['_TABLE']);
 
+                                if($recursiveDepth > 0) {
+                                    /** @var \TYPO3\CMS\Core\Database\QueryGenerator $queryGenerator */
+                                    $queryGenerator = GeneralUtility::makeInstance(\TYPO3\CMS\Core\Database\QueryGenerator::class);
```
tomasnorre: Any reason for the not using the BackendUtility::BEgetRootLine() ? Then we would not need the QueryGenerator

Classes/Controller/CrawlerController.php
```diff
@@ -810,25 +823,36 @@ public function expandParameters($paramArray, $pid)
...
+                                if(!empty($addTable)) {
+                                    // TODO: Check if this works as intended!
+                                    $queryBuilder->add('from', $addTable);
+                                }
                                 $transOrigPointerField = $GLOBALS['TCA'][$subpartParams['_TABLE']]['ctrl']['transOrigPointerField'];
 
                                 if ($subpartParams['_ENABLELANG'] && $transOrigPointerField) {
                                     $queryBuilder->andWhere(
                                         $queryBuilder->expr()->lte(
-                                            $queryBuilder->quoteIdentifier($transOrigPointerField),
```
tomasnorre: Why is this removed ?

Classes/Controller/CrawlerController.php
```diff
@@ -838,7 +862,7 @@ public function expandParameters($paramArray, $pid)
 
                                 $rows = [];
                                 while ($row = $statement->fetch()) {
-                                    $rows[$fieldName] = $row;
+                                    $rows[$row[$fieldName]] = $row;
```
tomasnorre: Looks like something, not related to your change, should be refactored here, as we have switch to a domain/model

Classes/Controller/CrawlerController.php
```diff
@@ -1229,13 +1253,15 @@ protected function getDuplicateRowsIfExist($tstamp, $fieldArray)
                 );
         }
 
-        $statement = $queryBuilder
-            ->andWhere('exec_time != 0')
-            ->andWhere('process_id != 0')
+        $queryBuilder
+            /* ->andWhere('exec_time != 0') # this was ' AND NOT exec_time' therfore functional test fails => change back */ ->andWhere('NOT exec_time')
```
tomasnorre: Remove if not needed.

Tests/Functional/Controller/CrawlerControllerTest.php
```diff
@@ -182,6 +192,300 @@ public function getLogEntriesForSetId($setId, $filter, $doFlush, $doFullFlush, $
...
+
+        unset($configuration['paramExpanded'], $configuration['URLs']);
+        $newCheckSum = md5(serialize($configuration));
+        $this->assertEquals(
+            $newCheckSum,
+            $this->subject->_call('getConfigurationHash', $configuration)
+        );
+    }
+
+// CrawlerController::isCrawlingProtocolHttps() removed in [7f7f07b]
```
tomasnorre: Please remove if not needed

tomasnorre:
You can do you adjustments now. Thanks

Classes/Controller/CrawlerController.php
vzz3: BackendUtility::BEgetRootLine() does a different thing. It returns the path from the provided PID back to the root page (PID=0). However, we need all PIDs of a sub tree.

Classes/Controller/CrawlerController.php
vzz3: because it result in a wrong SQL query. The column name gets "double" quoted.
e.g.:
```suggestion
 [ Doctrine\DBAL\Exception\InvalidFieldNameException ]                                                                           
  An exception occurred while executing 'SELECT `uid` FROM `tx_news_domain_model_news` WHERE (`pid` IN (?, ?, ?)) AND (hidden =   
  0) AND (```l10n_parent``` <= 0) AND (`tx_news_domain_model_news`.`deleted` = 0)' with params [46, 274, 275]:                    
                                                                                                                                  
  Unknown column '`l10n_parent`' in 'where clause'
```

Classes/Controller/CrawlerController.php
vzz3: The _TABLE based parameter expansion doesn't work at all with the old line
```
$rows[$fieldName] = $row;
```
because the array $rows can only contain a single row. The array item $fieldName gets overwritten in ever loop iteration. Therefore, I was forced to fix it, otherwise it would not have bin possible to test any of my changes.
Probably it is possible to just use
```
$rows[] = $row;
```
instead of 
```
$rows[$row[$fieldName]] = $row;
```
. However, my current code produces the exact same result (array) as the code of the crawler master branch. Which hopefully creates as least as possible side effects.

PR #486, opened 2019-12-01
[CI] Add Rector CI

TomasVotruba:
Based on @rectorphp [start contest](https://twitter.com/rectorphp/status/1196711403333332992) I send my first PR :) more are comming. Step by step.

**Feature:** adds first Rector run to CI.

The coding style was fixed somehow automatically.

The `"prefer-stable": true` in `composer.json` is best practise for development with dev dependencies (`"minimum-stability": "dev"`), so you always works with the same dependencies as the end user. It also speeds up composer install/update brutally.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -81,6 +81,7 @@ jobs:
     - env: TYPO3_VERSION=^10.0 COVERAGE=0 FUNCTIONAL=0
     - env: TYPO3_VERSION=dev-master COVERAGE=0 FUNCTIONAL=0
     - env: TYPO3_VERSION=dev-master COVERAGE=0 FUNCTIONAL=1
+    - name: "Rector CI"
   include:
     - stage: test
       env: TYPO3_VERSION=^9.5 COVERAGE=0 FUNCTIONAL=0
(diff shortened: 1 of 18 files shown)


TomasVotruba:
Not sure why scrutinizer fails.

Personally I recommend it to remove it for much faster feedback and local feedback [from PHPStan + Rector](https://www.tomasvotruba.cz/blog/2017/01/28/why-I-switched-scrutinizer-for-phpstan-and-you-should-too/)

tomasnorre:
Thanks for your PR. I will look into to the Scrutinizer. 

I've always like working with scrutinizer, so I need to look into alternatives before dropping/replacing it.

TomasVotruba:
That was fast!

Please do, I used it for 3 years before, but it didn't give any useful information and always kept failing.

tomasnorre:
I have seen after merging that the rector step failed as the TYPO3_VERSION wasn't defined.
I've then overwritten the before_step with `composer install` to have rector present, and which TYPO3 version is in this case not important, as it is static code analysis of the crawler itself. But as rector needs to be installed, this is done with composer install.

I get this error after that. 
https://travis-ci.org/AOEpeople/crawler/jobs/619259735

And not sure what causes this, have created a branch for testing:
https://github.com/AOEpeople/crawler/pull/488

TomasVotruba:
I'll fix in next PR.

Could you link me to error with constant? I didn't notice it

tomasnorre:
Sure: 
https://travis-ci.org/AOEpeople/crawler/jobs/619255699

Thanks, I'll close my PR then.

PR #488, opened 2019-12-01
[TASK] Add before step to rector

tomasnorre:


tomasnorre:
Not needed @TomasVotruba will fix it in next PR.

TomasVotruba:
I can see that now, pretty hidden

tomasnorre:
Yeah, I also just saw it by coincidence, it due to bad structure of the travis-ci.yml that one would oversee that part.

TomasVotruba:
We'll fix that :)

PR #489, opened 2019-12-01
Travis: cleanup and split jobs 

TomasVotruba:


TomasVotruba:
Not master again.

Btw could you make current dev branch as default? It would prevent such errors

PR #490, opened 2019-12-01
[Travis] Cleanup and split jobs clearly

TomasVotruba:


.travis.yml
```diff
@@ -15,21 +15,21 @@ addons:
     packages:
       - parallel
 
+env:
+  # default tested typo 3 version
```
tomasnorre: It's TYPO3 ;) But I'll change that :)

TomasVotruba: Sorry :+1:

TomasVotruba: This is WIP, do not merge yet

tomasnorre: ok. 

Please add WIP to the next PRs, and remove when ready for merge.

.travis.yml
```diff
@@ -15,21 +15,27 @@ addons:
...
 
+php:
+  - '7.2'
+  - '7.3'
+
+env:
+  # tested TYPO3 version
+  - TYPO3_VERSION=^9.5
+  - TYPO3_VERSION=^10.0
+  - TYPO3_VERSION=dev-master
```
TomasVotruba: These two add matrix 2x3 for both unit and functional tests.
Total battery of 2x3x2 = 12 sets.

Btw, I see 10.1 was released jsut 2 months ago: https://github.com/TYPO3/TYPO3.CMS/releases/tag/v10.1.0

So testing `dev-master` doesn't really add value here - that's rather responsibility of the `TYPO3/TYPO3.CMS` package.
If the 11 would be behind corner, that would more sense.

TomasVotruba:
I was about to check if this PR passes or what needs to be done here... and  CI passes, again scrutinizer is broken :/

tomasnorre:
Comment out the scrutinizer part, don't want to block you with that for now. Isn't mission critical in this branch atm.

.travis.yml
tomasnorre: The reason why I've added dev-master is because the `nimut/testing-framekwork` that we use for unit/functional tests isn't ^10.1 compatible (missing release) but runs with dev-master.

But honestly, I haven't look into it lately to see if this could be done better.

TomasVotruba:
Thank you!

Could you check matrix if it's ok for you? CI passes like before, I didn't modify any "allowed failures", so it's ready for me

tomasnorre:
Looks good to me. Do I get it correctly that it's ready for merge then ?

TomasVotruba:
:+1: Yes yes.

More PR is coming :)

PR #495, opened 2019-12-03
Apply static + PSR-12 set

TomasVotruba:
- apply `dead-code` Rector set
- add more powerful and easy to use [ECS](https://github.com/Symplify/EasyCodingStandard) over PHP-CS-Fixer
- apply [PSR-12](https://www.php-fig.org/psr/psr-12/) coding standard ([PSR-2 is deprecated](https://github.com/php-fig/fig-standards/commit/bc7741257a341146620a06675dd56b9fa8ebd882))


Classes/Api/CrawlerApi.php
```diff
@@ -1,4 +1,5 @@
 <?php
+
```
tomasnorre: Is this PSR-2 ? I have always preferred without spacing, but if it's the standard, we will stick to it :)

TomasVotruba: Updated description above :) see

TomasVotruba:
CI is passing, ready to merge

In next PR I'll try to add php70 and php71 upgrade.

Anything else you'd like me to check? 

## Standardize?

It's really confusing that source is spread all over the repository in chaos. It should be located in `/src`. That would make using composer and all CI tools much easier.

Also, is there some special reason for having weird `vendor` and `bin` locations? They could be removed as well

tomasnorre:
What do you mean by 
> In next PR I'll try to add php70 and php71 upgrade.

The folder structure cannot be changed, that is the standard of TYPO3 Extensions, so this has to stay. 

> Anything else you'd like me to check?

I don't really have the overview what could be done to improve, i'm learning every day too, and welcoming suggestions and help.

> Also, is there some special reason for having weird vendor and bin locations? They could be removed as well

The reason for having this, is to easier clean up after building etc. It's not standard in TYPO3, but a lot of people do it this way. 

https://github.com/AOEpeople/crawler/blob/typo3v9/.travis.yml#L41 is for instance to similar a Running TYPO3 for testing, as not all tests can run as standalone.

tomasnorre:
I'll do the code review later this evening.

Classes/Hooks/IndexedSearchCrawlerFilesHook.php
```diff
@@ -29,7 +30,7 @@ class IndexedSearchCrawlerFilesHook
      * @param object $pObj Parent object (tx_crawler lib)
```
tomasnorre: The Param need to removed in DocBlock as well.

Classes/Hooks/IndexedSearchCrawlerHook.php
```diff
@@ -835,7 +836,7 @@ public function deleteFromIndex($id)
      * @param mixed $value Target value (ignored)
```
tomasnorre: The two params should be removed from DocBlock as well.

Classes/Hooks/StaticFileCacheCreateUriHook.php
```diff
@@ -43,7 +44,7 @@ class StaticFileCacheCreateUriHook
      * @throws \Exception
      * @return void
      */
-    public function initialize(array $parameters, \SFC\NcStaticfilecache\StaticFileCache $parent)
+    public function initialize(array $parameters)
```
tomasnorre: I'm not sure this works, Think there is some magic behind it. I would need to check, as we added back in the second parameter some time ago.. 

Will need to check before I can accept the PR.

TomasVotruba:
I see. Thanks for explaining

composer.json
```diff
@@ -60,22 +58,13 @@
             "export TYPO3_PATH_WEB=$PWD/.Build/Web"
         ],
         "test:unit": [
-            "[ -e .Build/bin/phpunit ] || composer update",
```
tomasnorre: Why are these removed? It's just helper conditions that when the phpunit isn't there the composer update is executed. Handy for local development.

tomasnorre:
> In next PR I'll try to add php70 and php71 upgrade.

If you are referring to PHP 7.0 and 7.1 it can be skipped TYPO3 9. Support PHP 7.2+ so 7.0. and 7.1 would not work, so din't spent any time on that :)

composer.json
TomasVotruba: It's already in composer, right?

TomasVotruba:
> If you are referring to PHP 7.0 and 7.1 it can be skipped TYPO3 9. Support PHP 7.2+ so 7.0. and 7.1 would not work, so din't spent any time on that :)

Yes. To get to PHP 7.2, you can add features from 7.1 and 7.0, here is wip:
https://github.com/TomasVotruba/crawler-1/pull/1

composer.json
TomasVotruba: It should be in composer explicitly (now is missing and hoping to be installed via 3rd party).

When this gets merged, we can add it explictly here: https://github.com/Nimut/testing-framework/pull/129

.travis.yml
```diff
@@ -10,6 +10,10 @@ notifications:
 
 sudo: false
 
+before_install:
+  # turn off XDebug for speed up
+  - phpenv config-rm xdebug.ini || return 0
```
TomasVotruba: This will speedup every job, that does not require code coverage

tomasnorre: I know.. Just never did it, because i didn't saw a good way to do it.

composer.json
tomasnorre: Ja, it's in composer, but it's not always in a full system setup.. 

If I just checkout this extension, I would need to do a composer install/update to be able to run the tests and get the phpunit binary in this context too. 

That's why I have these "helpers". But If you see a better way I am listening.

TomasVotruba: You always need to run composer install to make package work anyway

tomasnorre: You're partly right in TYPO3 Context. I can do changes and push them to my development environment, and do manual testing.. and I could commit.

We have added this step, to ease and enable developers to test faster, as the script-path makes sure that the prerequisites are available if not already present. 

It's not disrespect of you suggestion, but would like to keep that part as is.

TomasVotruba: To make it right, you'd have to add it to every ci command. E.g. now phpstan and coding standard and rector won't work.

Anyhow. Feel free to merge it with revert of this commit. I'm currently busy

tomasnorre: No stress. I have problem with merging as is.. 
And get your point in inconsistency.. 

I Would still need to check a few things from the review first as well.

tomasnorre: But we will keep it as your suggest it, then I'll evaluate later.. So no need to change that.

Classes/Hooks/StaticFileCacheCreateUriHook.php
tomasnorre: I have reached out to the person Providing the extension we are hooking into.
https://twitter.com/tomasnorre/status/1202312054733979659

I think we might be able to drop this hook completely, but I'll follow up on this.

TomasVotruba: I'd recommend do that in separate PR, as it's rather more informed change. This PR only contains static-based changes

composer.json
TomasVotruba: I found some time and reverted as you needed. I realized I'm not TYPO3 expert and have just different habbits

TomasVotruba:
Travis CI passes :heavy_check_mark:

Issue #504, opened 2019-12-12
[BUG] declare(strict_types=1); breaks functionality


tomasnorre:
@infabo has made me aware of that the `declare(strict_types=1);` break functionality. 

E.g. from the https://github.com/aoepeople/crawler/blob/typo3v9/Classes/Backend/BackendModule.php#L1003
 
```
 protected function selectorBox($optArray, $name, $value, bool $multiple): string
    {
        $options = [];
        foreach ($optArray as $key => $val) {
            $selected = (!$multiple && !strcmp($value, $key)) || ($multiple && in_array($key, (array)$value));
            $options[] = '
				<option value="' . htmlspecialchars($key) . '"' . ($selected ? ' selected="selected"' : '') . '>' . htmlspecialchars($val) . '</option>';
        }

        return '<select class="form-control" name="' . htmlspecialchars($name . ($multiple ? '[]' : '')) . '"' . ($multiple ? ' multiple' : '') . '>' . implode('', $options) . '</select>';
    }
```

the `strcmp` expects strings for comparision, `$key` is `int`, the `htmlspecialchars()` expects a string again `$key` is `int`.

PR #505, opened 2019-12-13
[FIX] cast variable types to comply strict_types

infabo:
see #504


tomasnorre:
Thanks.

Do you want to take care of more of these changes as well? 
These can be merge, but there are more to be fixed.

I can also merge this as a first step in the right direction, and keep the issue open.

infabo:
These were the ones i stumpled upon.

1) add new scheduler task
2) buildqueue
3) processqueue

but I did not test all config variants or flushqueue or alike. But I can provide possible fixes when I encounter them.

PR #512, opened 2020-01-13
[BUGFIX] Load crawler initialization before page resolver

rengaw83:
The initialisation of the crawler has to be performed before the page resolver kicks in. otherwise the fe user groups could not be set and the page resolver responses a 403 for restricted pages.


tomasnorre:
Thanks for your fix, you can add your self to the Contributors.md if you like.

PR #526, opened 2020-03-03
[BUG FIX] Extensions throws Invalide argument exception with TYPO3 v9…

GhanshyamBhava:
….5.14

Extension throws a fatal error while depth is infinite level when you Start Crawling. 

`	
Oops, an error occurred!
Argument 2 passed to AOE\Crawler\Controller\CrawlerController::getConfigurationsForBranch() must be of the type integer, string given, called in /var/www/html/GNB/T3Demos/TYPO3.CMS.9/typo3conf/ext/crawler/Classes/Backend/BackendModule.php on line 429`

Configuration: https://prnt.sc/rav5xu
Error: https://prnt.sc/rav6k1

### DIFF
--- a/Classes/Backend/BackendModule.php
+++ b/Classes/Backend/BackendModule.php
@@ -426,7 +426,7 @@ class BackendModule
         );
 
         // Configurations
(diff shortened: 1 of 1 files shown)


tomasnorre:
Thanks, are you on the current dev-typo3v9 branch? 

I'll verify this and have it merged if it solves the problem. Thanks.

PR #527, opened 2020-03-06
Separate middleware initialization

IchHabRecht:
The frontend usergroup needs to be set before the typo3/cms-frontend/page-resolver is processed. At that time the TSFE controller isn't initialized yet, so the TSFE configuration needs to be done at a later point.


tomasnorre:
Thanks a lot, will have a look. 
Could you perhaps run the `.Build/bin/ecs check Classes Configuration Tests --fix` on your changes and push again :)

Issue #481, opened 2019-11-28
[BUG] Direct Requests doesn't work


tomasnorre:
Steps to reproduce:

*Current Behaviour*
1. Go to the Crawler Log
2. Press the "process wheel"
3. Status column isn't updated

*Expected Behaviour*
The Status column should be updated with a ok or an info-text on the error.

![crawler-log](https://user-images.githubusercontent.com/1212481/69781983-0f7d2f00-11b0-11ea-875b-f32f18796d05.png)

PR #534, opened 2020-04-07
[BUGFIX] Check if indexes are set before using the in bootstrap.php (#481)

tomasnorre:
Resolves: #481


tomasnorre:
Unfortunately it doesn't solve the problem, have tested on a different instance than my devbox.

PR #553, opened 2020-04-26
[cs] add common set

TomasVotruba:
While I was at upgrading frienly package https://github.com/sabbelasichon/typo3-rector/pull/1404, I've noticed ecs is missing few of basic sets here.

Found a bit time to fix that :)

### DIFF
--- a/Classes/Api/CrawlerApi.php
+++ b/Classes/Api/CrawlerApi.php
@@ -115,27 +115,6 @@ class CrawlerApi
     }
 
     /**
-     * Method to get an instance of the internal crawler singleton
-     *
-     * @return CrawlerController Instance of the crawler lib
-     *
-     * @throws \Exception
-     */
-    protected function findCrawler()
-    {
-        if (! is_object($this->crawlerController)) {
-            $this->crawlerController = GeneralUtility::makeInstance(CrawlerController::class);
-            $this->crawlerController->setID = GeneralUtility::md5int(microtime());
-        }
-
-        if (is_object($this->crawlerController)) {
-            return $this->crawlerController;
-        } else {
-            throw new \Exception('no crawler object', 1512659759);
-        }
-    }
-
-    /**
      * Adds a page to the crawlerqueue by uid
      *
      * @param int $uid uid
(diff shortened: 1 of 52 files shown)


Classes/Api/CrawlerApi.php
```diff
@@ -150,24 +126,6 @@ public function addPageToQueue($uid): void
...
-            foreach ($configurations as $confKey => $confArray) {
-                if (!in_array($confKey, $this->allowedConfigurations)) {
-                    unset($configurations[$confKey]);
-                }
-            }
-        }
-
-        return $configurations;
-    }
-
```
TomasVotruba, line 170: Methods are now sorted by `public`, `protected`, `private`, that's why many of them are moved.

tomasnorre:
Thanks a lot, have quite a number of branches open atm, so would need to be a little careful with merging. 

But will check, thanks for taking your time, not only here, but also in typo3-rector by Sebastian.

Tests/Acceptance/Support/BackendTester.php
```diff
@@ -39,7 +39,6 @@ class BackendTester extends \Codeception\Actor
 {
     use BackendTesterActions;
     use FrameSteps;
```
tomasnorre, line 41: Which rule triggers this? 
I would prefer an empty line between use in class context, and not have it followed directly by a comment.

tomasnorre:
Would you consider making the request only for the: Sets, Composer and .editorconfig? 
Then I'll easily rebase and apply.

Tests/Acceptance/Support/BackendTester.php
TomasVotruba: It was caused by Slevomat bug, that doesn't take account follow up of comment only. It was just in this file. I've ignored the rule and restored this space.

All the other trait + content are separated by single space correctly.

image

TomasVotruba:
> have quite a number of branches open atm

That's what I love about coding standards, that this doesn't matter. 

After merge, all you need to do is update `ecs.yaml` only and run:

```bash
.Build/bin/ecs check --fix
```

on particular branch before merge. This prevents rebase conflicts and updated code to new style :+1:

Tests/Acceptance/Support/BackendTester.php
tomasnorre: Ja, that I saw, so was considering why it was like this in this case. 

Thanks.

TomasVotruba:
Woohoo, that was fast :)

:+1: Enjoy

TomasVotruba:
In case of any troubles, let me know

tomasnorre:
Thanks, I like not having contributors waiting. Otherwise they will lose motivation at some point.

TomasVotruba:
Exactly :+1: 

I can do 10 PRs a day, but only 1 in 2 months.

PR #555, opened 2020-04-29
[FIX] get fe_user from $GLOBALS['TSFE']

werkraum-admin:
Setting the $frontendUser with $request->getAttribute('frontend.user') resulted in a \stdClass

Since this middleware is called after the core fe_user authenticator we can assume an instance of FrontendUserAuthentication is set in $GLOBALS['TSFE']->fe_user


tomasnorre:
Thanks for your pr @werkraum-admin,

In which TYPO3 version do you have an issue with this?

bmack:
Yes. You are right: 👍

PR #556, opened 2020-04-29
Fix/backend controller

werkraum-admin:
In case of a PDF Document on a page $parameters['url'] was not set -> fallback to 'alturl'. Otherwise the backend throws an error.

The same Issue accoured with getResFeVars


tomasnorre:
Thanks for our PR @kozodev

PR #562, opened 2020-05-02
[TASK] Add and optimize some behavior tests

tomasnorre:


TomasVotruba:
I'll try to run this PR locally and see what goes wrong with Rector

TomasVotruba:
I might have an idea... [`PropertyProperty` is actually subnode of `Property`](https://github.com/rectorphp/rector/blob/master/docs/nodes_overview.md#phpparsernodestmtproperty). 

So maybe the docblock is detected on wrong node. Will investigate

TomasVotruba:
Of course that's it :) 

Would you like to send PR, since you've invested a lot into issue report?

It's pretty simple, just update this class: https://github.com/rectorphp/rector/blob/master/src/Exclusion/Check/ExcludeByDocBlockExclusionCheck.php#L15

```diff
+use PhpParser\Node\Const_;
+use PhpParser\Node\Stmt\PropertyProperty;

// ...

public function isNodeSkippedByRector(PhpRectorInterface $phpRector, Node $node): bool
    {
+       if ($node instanceof PropertyProperty || $node instanceof Const_) {
+           $node = $node->getAttribute(AttributeKey::PARENT_NODE);
+       }

        $comment = $node->getDocComment();
        if ($comment !== null && $this->checkCommentForIgnore($phpRector, $comment)) {
            return true;
        }
```

tomasnorre:
I'll check and send PR.

tomasnorre:
It's working but the tests `composer complete-check` isn't green. 

```
$ vendor/bin/phpstan analyse --ansi --error-format symplify

 --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 src/Exclusion/Check/ExcludeByDocBlockExclusionCheck.php:24
 --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 - '#Cannot call method getDocComment\(\) on PhpParser\\Node\|null#'
 --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

 --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 src/Exclusion/Check/ExcludeByDocBlockExclusionCheck.php:30
 --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 - '#Cannot call method getAttribute\(\) on PhpParser\\Node\|null#'
 --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

The problem as I see it is that the 
```
 if ($node instanceof PropertyProperty || $node instanceof Const_) {
            $node = $node->getAttribute(AttributeKey::PARENT_NODE);
        }
```

resolves in a different node class or null where the `->getComment()` and `->getAttribute()` isn't available.

TomasVotruba:
What do you suggest?

tomasnorre:
I'll have to consider that to have the code as clean as possible. :)

tomasnorre:
```php
 if ($node instanceof PropertyProperty || $node instanceof Const_) {
            $node = $node->getAttribute(AttributeKey::PARENT_NODE);
            if ($node === null) {
                return false;
            }
        }
```

This would be my suggestion. By will not commit before I hear your comments.

TomasVotruba:
Exactly my thoughts.

Issue #417, opened 2019-08-21
Replace all serialize/unserialize with json_encode/decode where applicable


bmack:
We should get rid of serialize/unserialize and replace that with JSON data for the database fields, if there are no objects stored in the DB.

tomasnorre:
Resolved with #564

PR #564, opened 2020-05-04
[WIP] json_encode instead of serialize data before writte to db.

tomasnorre:
Resolves issue #417

### DIFF
--- a/Classes/Converter/JsonCompatibilityConverter.php
+++ b/Classes/Converter/JsonCompatibilityConverter.php
@@ -32,7 +32,7 @@ class JsonCompatibilityConverter
      */
     public function convert(string $dataString)
     {
-        $unserialized = unserialize($dataString);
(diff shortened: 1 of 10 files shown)


Classes/Converter/JsonCompatibilityConverter.php
```diff
@@ -0,0 +1,47 @@
...
+     * in the database. To ensure that older crawler entries, which have already been stored as serialized data
+     * still works, we have added this converter that can be used for the reading part. The writing part will be done
+     * in json from now on.
+     * @see https://github.com/AOEpeople/crawler/issues/417
+     *
+     * @return array|bool
+     */
+    public function convert(string $dataString)
+    {
+        $unserialized = unserialize($dataString);
```
bmack, line 35: ideally - since we know what people are storing in their database fields, we could use `unserialize($dataString, ['allowed_classes' => false]);`

tomasnorre:
Anything else @bmack?

Issue #545, opened 2020-04-16
module:info htmlspecialchars() NULL given - for file entries


TrueType:
## Bug Report
module:info -> SiteCrawler -> Crawler-Log

**Current Behavior**
Core: Exception handler (WEB): Uncaught TYPO3 Exception: htmlspecialchars() expects parameter 1 to be string, null given | TypeError thrown in file /var/www/html/../current/webroot/web/typo3conf/ext/crawler/Classes/Backend/BackendModule.php in line 659. Requested URL: http:/mydomain.tld/typo3/index.php?route=%%2Fweb%%2Finfo%%2F&token=--AnonymizedToken--&id=1 

**Expected behavior/output**
the $parameter['url'] is not set for files.

**Environment**
- Crawler version(s): [e.g. 9.0.0]
- TYPO3 version(s): [e.g. 9.5.15]
- Is your TYPO3 installation set up with Composer (Composer Mode): [yes]

tomasnorre:
Do you always get this error when going to the Crawler Log and did you do something before the error happend?

TrueType:
As I mentioned the problem occurs for files only. As you process pages, which contain links to files, for each file an additional entry is put into the queue. At the point in code mentioned serialized data is processed for every entry and for files there is not "url" index given. There is only a path on the server.
To reproduce: create a page with file reference. Get the page in the queue. Index it by processing the queue. It should create an entry in the queue for the file found while indexing the page. I guess that is enough to crash the Crawler Log already.

tomasnorre:
Thanks for the update, I will check.
Edited the text to make it a little easier to read.

tomasnorre:
Hi @TrueType,

I cannot reproduce this. Do I do anything different than want to meant? 
Please see video linked.
https://www.dropbox.com/s/k8w9anpvl3cbzrd/issue-545.mp4?dl=0

TrueType:
Hi Tomas,

sry, for coming back to you that late.

I checked your video. Makes me wonder why your processing does to work on the pdf file? As I come back from starting a process I see two entries. One is for the page, which is on status OK, one is newly created - not yet processed - and is for the file found while processing the page. 

So your system does not yet process files. 

You may check for the settings on EXT:indexed_search:

`basic.useCrawlerForExternalFiles = 1`

And the binary to handle pdf files

`/usr/bin/pdftotext`

tomasnorre:
[...]

PR #567, opened 2020-05-07
module:info prevent htmlspecialchars() error with file entries

TrueType:
#545 
major PR ;)


TrueType:
total bullsh**. You fixed that already. Working on customers systems where I can hardly type.

tomasnorre:
I fixed it diffent though, or someone else did the other day.. I cannot take credit for that. 

But this solve you issue with the PDF? Or is that still open ?

Issue #587, opened 2020-06-05
Crawler exits with an error


kusiboss:
## Bug Report

**Current Behavior**
Crawler exits with an error when trying to fetch certain pages. 
	
`Call to a member function getStatusCode() on null`

**Expected behavior/output**
Crawler should continue and log the error.

**Environment**
- Crawler version: 9.5.14
- TYPO3 version: 9.0.1
- Composer mode: yes

**Possible Solution**
It happens when a request doesn't return a status code because for example there is a redirect loop.

tomasnorre:
Hi,

Thanks for your information. Redirect loops isn't wanted regardless of crawler or not.  I would suggest you solve that first, and test if you still have an issue.

If still an issue, please let me know and add more description on how to reproduce this.

kusiboss:
Hi,
In my opinion the crawler shouldn't stop in any case when there a problem accessing a page. Redirects are made by authors and mistakes can happen. The error doesn't happen only with redirect loops, it happens when a request doesn't return a status code.

class GuzzleExecutionStrategy
```
catch (RequestException $e) {
            $this->logger->debug(
                sprintf('Error while opening "%s" - ' . $e->getResponse()->getStatusCode() . chr(32) . $e->getResponse()->getReasonPhrase(), $url),
                ['crawlerId' => $crawlerId]
            );
            return $e->getResponse()->getStatusCode() . chr(32) . $e->getResponse()->getReasonPhrase();
        }
```
The catch part shouldn't throw an exception by itself.

Reproduce:
Add a redirect with target to itself (redirect loop). Start crawling. Crawler will stop as soon as it tries to access that page.

tomasnorre:
Thanks for the additional information, I'll have a look at this.

brotkrueml:
See #584, I will provide a PR shortly.

tomasnorre:
I was thinking that it could be related but wasn't sure.

Issue #584, opened 2020-06-02
buildQueue creates entry for hidden/non-existing localised pages


brotkrueml:
## Bug Report

**Current Behavior**
Given is a single-tree multi-lingual project with en als default language and de, pl and tr as other languages. Now create a configuration record with "pidsonly" for a page where only en and de exists as available pages, pl is hidden and tr is not existing. When calling crawler:buildQueue for this configuration four entries are created into the queue table, one for each of the four languages.

This also happens for the default language, when "Hide default language" is activated in the page properties.

As invalid pages are added to the page the processQueue exits with an error when processing a non-existing page.

**Expected behavior/output**
Only rows for the available translated pages should be created in the queue table.

**Environment**
- Crawler version: 9.0.1
- TYPO3 version: 10.4.3
- Is your TYPO3 installation set up with Composer (Composer Mode): yes

tomasnorre:
Hi @brotkrueml 

I understand the problem here and it's a bug in the crawler, unknown till now. 
Do you agree with me, that it's not "critical" as the existing pages will still be crawler as they should be? 

Please comment on this so that I know if I assigned the correct priority to this bug.

brotkrueml:
I want to use the extension as cache warmup. Every page that does not exist means, the process command exits with an error and the next run of the process command (when the given grace time is over) will crawl some other pages till the next error. And there are plenty of pages which are not translated in this specific project.

tomasnorre:
Then it's a higher priority than I saw this. I'll change the label to have it addressed earlier. 
Didn't expect the process to stop, with the 9.0.0 version, it started handling 404 and 50x better, but sounds like there still is a problem.

tomasnorre:
I've tried to reproduce this, but on a 9LTS, which I cannot. You can see that all three languages (default, da, de) are added to the queue, and no errors by processing. 

![Screenshot from 2020-06-03 17:19:11](https://user-images.githubusercontent.com/1212481/83655352-9deac700-a5be-11ea-90b1-cc3339b008da.png)
![Screenshot from 2020-06-03 17:19:33](https://user-images.githubusercontent.com/1212481/83655355-9e835d80-a5be-11ea-80c6-dbc3f513b122.png)

I'll try to update my devbox to 10 LTS to see if it makes any difference to me. 

FYI: I totally agree with you that they should be added to the queue in the first place, just want to figure out why it gives you problems first.

tomasnorre:
[...]

PR #588, opened 2020-06-05
[BUGFIX] Handle empty response when fetching URL

brotkrueml:
This case can occur, when the connection is resetted from the server.

Resolves: #587
Related: #584

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog TYPO3 Crawler
 
+### Changed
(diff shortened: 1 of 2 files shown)


tomasnorre:
Could you please add a line to the CHANGELOG.md? then I'll be happy to merge.

tomasnorre:
I'll merge this when the travis is done, but I'll change it to a FIXED as it's a bug. 
But I can take care of that, just really happy that you are helping out.

PR #589, opened 2020-06-05
[TASK] List specific doktypes for skipping page

brotkrueml:
With TYPO3 10.4 the limitation of custom doktypes to a number less
than 200 was lifted. This patch makes it possible to crawl on custom
doktypes with a number greater than 200.

See also:
- https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/10.4/Important-18079-PagesdoktypeRestrictionForFrontendQueriesRefined.html
- https://github.com/TYPO3/TYPO3.CMS/blob/master/typo3/sysext/core/Classes/Domain/Repository/PageRepository.php#L103

### DIFF
--- a/Classes/Controller/CrawlerController.php
+++ b/Classes/Controller/CrawlerController.php
@@ -338,7 +338,7 @@ class CrawlerController implements LoggerAwareInterface
         }
 
         if (! $skipPage) {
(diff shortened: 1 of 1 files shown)


tomasnorre:
Thanks again, could I ask you to update the CHANGELOG.md here as well? will be happy to merge.

tomasnorre:
I've updated the CHANGELOG.md 

Thanks for your contribution.

PR #590, opened 2020-06-05
[TASK] Introduce .gitattributes

brotkrueml:
Files and folders which are defined as export-ignore are not
included in an archive and therefore not installed via composer
on production systems.


tomasnorre:
Thanks, I  wasn't aware of this. I've always cleaned up before doing a release, so therefore they shouldn't be in the artefact at all. But that's of course only in the TER release.

PR #591, opened 2020-06-09
[TASK] Correct typo in XML tag

brotkrueml:


brotkrueml:
The tests also fail locally for master. Seems to be another issue as far as I can see.

tomasnorre:
Yeah, I added some testes yesterday and didn't do the mocking on the timestamp `$GLOBALS ['exec_time']` properly. :(  Will look  into this.

tomasnorre:
I've commented out the tests, as I don't have the time to fix the timezone/datetime mocking now, and will not manage it before my vacation i'm affraid.

PR #592, opened 2020-06-09
[TASK] Add description to commands

brotkrueml:
The descriptions are shown in the overview of available commands when
vendor/bin/typo3 is called without any parameter.


brotkrueml:
Before:
crawler-commands-before

After:
crawler-commands-after

PR #598, opened 2020-06-21
[WIP] Use dependency injection in FlushQueueCommand

brotkrueml:

PR #599, opened 2020-06-24
[TASK] Allow symfony/console 5

CDRO:


CDRO:
@tomasnorre I think this change should not impact the project too much while also making it compatible with the latest version of `symfony/*` packages

tomasnorre:
Could you please update the CHANGELOG.md as well to reflect the new change? 
Thanks. Then I'll merge it.

CDRO:
Thanks for the feedback, I've updated the changelog.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Added
 * Update specific doktypes for skipping page [@brotkrueml](https://github.com/brotkrueml)
 * Tests for BackendModule
+* Support for TYPO3 10.4
```
tomasnorre: Was this blocking 10.4? I don't get why this has worked till now for the tests then. 

I would have expected a change log saying (Support for symfony/console 5.0 instead. :)

CDRO: Yeah, as you see in my follow-up commit i changed the line to Symfony console :-)

I guess Monday morning is too soon to write any usable Changelog ^^'

brotkrueml:
I tested symfony/console ^5 also some weeks ago, but had the problem, that in one case no integer value was returned (and so an error was thrown). With version 5 always a return value has to be given. I wanted to check it out why, but first add some tests to the commands.

CHANGELOG.md
tomasnorre: :) is all ok.. I've back from 2 weeks of vacation, so not my best state either.

CDRO:
Thanks @brotkrueml, I found the place and added the expected handling, but feel free to write some tests for it :-)

tomasnorre:
@CDRO @brotkrueml There are tests for the CommandControllers, they are just not included in the Code coverage yet, which isn't a big deal. 

It would be great if tests could be added to prove this has been solved, before added the change. @brotkrueml thanks for your feedback didn't have that on my mind, and would probably just have merged it :(

CDRO:
I'm wondering on how I could write a test to ensure that the return value is always an integer? Would a static code analysing test be sufficient here or what would we usually write in this case, as it isn't a functional nor a unit test IMHO.

tomasnorre:
I honestly don't know. 

My first thought was: 
If it's all type-safe, we should be quite sure we will always get an int returned. But if we have an edge case where it's not, it will result in a fatal error.

CDRO:
I guess the edge cases would be covered by the IDE when doing static analysis, maybe we can solve this by adding phpstan to the checks rather than have a test for it?

brotkrueml:
I started to unit test the commands, beginning with the FlushQueueCommand (which is the easiest): #598.

This way, we can test only the command itself (with according mocks to CrawlerController and QueryBuilder). If every path is covered, this would be sufficient.

Additionally, we can add "int" as return type to the execute() method, or as @return annotation. So IDEs will complain if there will be no int as return.

tomasnorre:
Sounds like an approach to me. But as mentioned, there are tests present already, so don't know which more is needed, or if the "just" need to be extended.

https://github.com/AOEpeople/crawler/tree/master/Tests/Functional/Command

brotkrueml:
Yes, there are functional tests available, which test the commands as whole. But: Unit tests are easier to maintain, are cheaper and you can check edge cases better. You don't have to fill some wrong data which throws an exception to test that path (like in ProcessQueueCommand), just mock it and that's it.

So, in the future you can perhaps decouple the testing of the command with the testing of the CrawlerController (which is to monolithic and should be split up to separate classes in the long run - but that's another story ;-) ). So I created the #598 PR as a PoC to start with that. But that's just a suggestion and should be discussed in the PR.

I think the way to go for now would be to add "@inheritdoc" annotation, "@return int" annotation or "int" return type to the command. So IDEs will know about the type.

> Sounds like an approach to me. But as mentioned, there are tests present already, so don't know which more is needed, or if the "just" need to be extended.
> 
> https://github.com/AOEpeople/crawler/tree/master/Tests/Functional/Command

tomasnorre:
Sounds good to me.. I'm totally with you on the unit tests vs. functional tests.

tomasnorre:
Should we merge as is for now? after rebase. Or what do you think? 

@CDRO @brotkrueml

CDRO:
I think that we can merge this, since the PHPDoc is already set and "symfony/console"(https://github.com/symfony/console/blob/master/Command/Command.php#L162) does not define a return type yet.

tomasnorre:
@CDRO would you be so kind to rebase? Then I'll have it merged

tomasnorre:
I've resolved the conflict via Github wasn't a big one ;) 
I'll merge when the pipeline is done.

CDRO:
Oh, i rebased it now XD

PR #601, opened 2020-07-03
[DOCS] Add example for pageVeto hook

brotkrueml:
## Description

Add an example for the pageVeto hook.

Additionally:
- Name files where to register the hooks
- Use namespaced class for registration of pageVeto hook

**I have**

### DIFF
--- a/Documentation/Features/Hooks/Index.rst
+++ b/Documentation/Features/Hooks/Index.rst
@@ -5,6 +5,8 @@
 Hooks
 =====
 
+Register the following hooks in :file:`ext_localconf.php` of your extension.
+
 excludeDoktype Hook
 ===================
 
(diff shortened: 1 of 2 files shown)


Documentation/Features/Hooks/Index.rst
```diff
@@ -5,6 +5,9 @@
 Hooks
 =====
 
+Register the following hooks in :file:`typo3conf/LocalConfiguration.php`
+or :file:`typo3conf/AdditionalConfiguration.php`.
+
```
tomasnorre, line 10: I never registered hooks in the `LocalConfiguration.php` or `AdditionalConfiguration.php` I would always do it the `ext_localconf.php` for the extension holding the implementation. 

Is there a reason for you to suggest it like this that I'm not familiar with?

brotkrueml: Hmm, you are right. I will adjust it, good that we spoke about that :-)

PR #605, opened 2020-07-09
[TASK] Add resname attribute to trans-unit tags in XLF files

brotkrueml:
## Description

Crowdin doesn't show the information stored in the attribute "id".
If it should be visible, the attribute must be duplicated using
"resname".

find Resources/Private/Language/ -name '*.xlf' -exec sed -i -e \
    '/resname/!s/\(<trans-unit id="\)\([^"]*\)"/\1\2" resname="\2"/' \
    '{}' +

Additionally, the attribute "original" is filled with the path of the translation
file instead of dummy value "messages". This string is also shown in Crowdin
and helps users while translating.

**I have**


brotkrueml:
It the looks like this, which helps a lot to understand the context while translating:
image

PR #609, opened 2020-07-13
Add working test for DataHandlerHook

icanhazstring:
Add working addFlushedPages test

### DIFF
--- a/Tests/Unit/Hooks/DataHandlerHookTest.php
+++ b/Tests/Unit/Hooks/DataHandlerHookTest.php
@@ -1,0 +1,51 @@
+<?php
+
+declare(strict_types=1);
+
+namespace AOE\Crawler\Tests\Unit\Hooks;
+
+use AOE\Crawler\Api\CrawlerApi;
+use AOE\Crawler\Domain\Repository\QueueRepository;
+use AOE\Crawler\Hooks\DataHandlerHook;
+use Nimut\TestingFramework\TestCase\UnitTestCase;
+use TYPO3\CMS\Core\Authentication\BackendUserAuthentication;
+use TYPO3\CMS\Core\DataHandling\DataHandler;
+use TYPO3\CMS\Core\Utility\GeneralUtility;
+use TYPO3\CMS\Extbase\Object\ObjectManager;
+
+class DataHandlerHookTest extends UnitTestCase
+{
+    /**
+     * @test
+     */
+    public function itShouldAddPageToQueue(): void
+    {
+        $dataHandlerHook = new DataHandlerHook();
+
+        $crawlerApi = $this->prophesize(CrawlerApi::class);
+        $crawlerApi->addPageToQueue(1)->shouldBeCalled();
+        $queueRepository = $this->prophesize(QueueRepository::class);
+        $queueRepository->isPageInQueue(1)->willReturn(false);
+        $queueRepository->isPageInQueue(2)->willReturn(true);
+
+        $objectManager = $this->prophesize(ObjectManager::class);
+        $objectManager->get(QueueRepository::class)->willReturn($queueRepository->reveal());
+        $objectManager->get(CrawlerApi::class)->willReturn($crawlerApi->reveal());
+
+        GeneralUtility::addInstance(ObjectManager::class, $objectManager->reveal());
+
+        $beUser = new BackendUserAuthentication();
+        $beUser->workspace = 1;
+
+        $dataHandler = new DataHandler();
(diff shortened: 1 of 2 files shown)


Tests/Unit/Hooks/IndexedSearchCrawlerHookTest.php
```diff
@@ -2,7 +2,7 @@
 
 declare(strict_types=1);
 
-namespace AOE\Crawler\Tests\Unit\Hook;
+namespace AOE\Crawler\Tests\Unit\Hooks;
```
icanhazstring, line 5: Just rename to reflect the actual namespace under `Classes`

Tests/Unit/Hooks/DataHandlerHookTest.php
```diff
@@ -0,0 +1,61 @@
...
+     * Page with ID 1 is not queue, should be added
+     * Page with ID 2 is already in queue. Should NOT be added.
+     *
+     * @test
+     */
+    public function itShouldAddPageToQueue(): void
+    {
+        $dataHandlerHook = new DataHandlerHook();
+
+        $crawlerApi = $this->prophesize(CrawlerApi::class);
```
icanhazstring, line 31: `$crawlerApi` and `$queueRepository` are direct dependencies of your `DataHandlerHook` received through `GeneralUtility` and the `ObjectManager`.

You can now check on the case you want to test.

Tests/Unit/Hooks/DataHandlerHookTest.php
```diff
@@ -0,0 +1,61 @@
...
+        $queueRepository->isPageInQueue(2)->willReturn(true);
+
+        $objectManager = $this->prophesize(ObjectManager::class);
+        $objectManager->get(QueueRepository::class)->willReturn($queueRepository->reveal());
+        $objectManager->get(CrawlerApi::class)->willReturn($crawlerApi->reveal());
+
+        $cacheManager = $this->prophesize(CacheManager::class);
+        $cacheManager->getCache(Argument::any())->willReturn($this->prophesize(FrontendInterface::class)->reveal());
+
+        GeneralUtility::setSingletonInstance(CacheManager::class, $cacheManager->reveal());
```
icanhazstring, line 44: Seems like you can stub/mock these Objects quite well using `GeneraltUtility::addInstance()` and `GeneraltUtility::setSingletonInstance`.

Tests/Unit/Hooks/DataHandlerHookTest.php
```diff
@@ -0,0 +1,61 @@
...
+        GeneralUtility::setSingletonInstance(ObjectManager::class, $objectManager->reveal());
+
+        $beUser = new BackendUserAuthentication();
+        $beUser->workspace = 1;
+
+        $dataHandler = new DataHandler();
+        $cmd = ['pages' => [123 => ['version' => ['action' => 'swap']]]];
```
icanhazstring, line 51: At the moment the `DataHandlerHook` has much logic buried inside of it.
Every `private` method could be extracted into another class where you can simply keep these methods public and test them as well.

If you are not doing this, you tests for the `DataHandlerHook` might be tricky as you need to add permutations of you test for every outcome of every private method inside of it.

tomasnorre: But if the method is only used ones, and only in this class, why make a new class and make it public? 
Is this some pattern I don't know of? Or is it "only" to make testing easier.

icanhazstring: Its mostly to make testing easier.
It's also totally legit to have a class with one method.

In our DDD we have defined `Rules` these rules are merely an `if` wrapped in a class.
So for example inside your `addFlushedPagesTo...` method, you could have a special rule which wraps the `if` in the first line.

Every other `private` method could do the same. Also you could wrap all those private methods in there.
This way you will have all sorts of small and easy to understand unit tests instead of one big test case where its hard to understand what is happening.

See your unit tests as the documentation for each developer. Some will start digging into your code by reading your unit tests. If these are hard to understand it is hard to contribute.

tomasnorre: That point I get, but your first comment you said move private functions to a new class. 

The abstraction here, in your example, I totally agree with.

tomasnorre:
Will merge your changes, to have a better look at it, and to see why the tests are failing for TYPO3 10.

tomasnorre:
The tests were due to a composer dependency I added for testing, but didn't fit the version 10.. Now I does.

PR #614, opened 2020-08-05
Added Widget for displaying queue size

PhilippKuhlmay:
## Description

This PR adds a widget to the new EXT:dashboard.

No

**I have**

- [YES] Checked that CGL are followed
- [NO] Checked that the Tests are still working
- [YES] Added description to CHANGELOG.md (github-handle is optional)
- [NO] Added tests for the new code


PhilippKuhlmay:
This is still in Proof of Concept.

Classes/Widgets/Provider/QueueSizeDataProvider.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+declare(strict_types=1);
+
+namespace AOE\Crawler\Widgets\Provider;
+
```
tomasnorre, line 6: Please add license text, this applies to all PHP files

```
/*
 * (c) 2020 AOE GmbH <dev@aoe.com>
 *
 * This file is part of the TYPO3 Crawler Extension.
 *
 * It is free software; you can redistribute it and/or modify it under
 * the terms of the GNU General Public License, either version 2
 * of the License, or any later version.
 *
 * For the full copyright and license information, please read the
 * LICENSE.txt file that was distributed with this source code.
 *
 * The TYPO3 project - inspiring people to share!
 */
```

ext_emconf.php
```diff
@@ -3,19 +3,21 @@
     'title' => 'Site Crawler',
     'description' => 'Libraries and scripts for crawling the TYPO3 page tree.',
     'category' => 'module',
-    'state' => 'stable',
+    'state' => 'beta',
```
tomasnorre: This you cannot change as it's refering to the crawler and not the feature.

ext_emconf.php
```diff
@@ -3,19 +3,21 @@
...
     'uploadfolder' => 0,
     'createDirs' => '',
     'clearCacheOnLoad' => 0,
     'author' => 'Kasper Skaarhoj, Daniel Poetzinger, Fabrizio Branca, Tolleiv Nietsch, Timo Schmidt, Michael Klapper, Stefan Rotsch, Tomas Norre Mikkelsen, Tizian Schmidlin',
     'author_email' => 'dev@aoe.com',
     'author_company' => 'AOE GmbH',
-    'version' => '9.1.0',
+    'version' => '9.1.1',
```
tomasnorre: Please don't update versions here, just leave it as it is, we do this as part of a release, and technically as it's a feature , it would be a 9.2.0 :)

ext_emconf.php
PhilippKuhlmay: I changed it back. It was a mistake and shouldn't be commited.

ext_emconf.php
PhilippKuhlmay: I changed it back. It was a mistake and shouldn't be commited.

Classes/Widgets/Provider/QueueSizeDataProvider.php
PhilippKuhlmay: I added this to the php files

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog TYPO3 Crawler
 
+### Added
+* Adds a widget to the dashboard, which shows the size of the queue.
```
brotkrueml, line 4: Nitpicking: As "Added" is the header, the following text would be nicer: "Dashboard widget, which shows the size of the queue"

Classes/Widgets/Provider/QueueSizeDataProvider.php
```diff
@@ -0,0 +1,38 @@
...
+use TYPO3\CMS\Core\Utility\GeneralUtility;
+use TYPO3\CMS\Dashboard\Widgets\NumberWithIconDataProviderInterface;
+use TYPO3\CMS\Extbase\Object\ObjectManager;
+
+class QueueSizeDataProvider implements NumberWithIconDataProviderInterface
+{
+    public function getNumber(): int
+    {
+        $objectManager = GeneralUtility::makeInstance(ObjectManager::class);
+        $queueRepository = $objectManager->get(QueueRepository::class);
```
brotkrueml, line 32: ObjectManager->get() is deprecated in v10. But you can use Dependency Injection to get the QueueRepository easily. So this method is then a one-liner.

Example: https://github.com/brotkrueml/typo3-jobrouter-process/blob/master/Classes/Dashboard/Provider/TransferTypeChartDataProvider.php#L39

Configuration/Services.php
```diff
@@ -0,0 +1,52 @@
...
+use AOE\Crawler\Widgets\QueueSizeWidget;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
+use Symfony\Component\DependencyInjection\Reference;
+use TYPO3\CMS\Core\Utility\ExtensionManagementUtility;
+
+return function (ContainerConfigurator $configurator, ContainerBuilder $containerBuilder) {
+    $services = $configurator->services();
+
+    if (ExtensionManagementUtility::isLoaded('reports')) {
```
brotkrueml, line 32: Instead of "reports" this must be "dashboard". If the dashboard extension is not available the DI cache cannot be build. Please test also without installed dashboard extension.

Configuration/Services.yaml
```diff
@@ -21,3 +21,21 @@ services:
         tags:
             - name: 'console.command'
               command: 'crawler:processQueue'
+
+    dashboard.widget.crawler.queuesize:
```
brotkrueml, line 25: AFAIK, this must also be in Services.yaml. Background: QueueSizeDataProvider implements NumberWithIconDataProviderInterface which is part of dashboard ext.

PS: Is this a leftover? It's the same in Services.php?

Configuration/Services.php
```diff
@@ -0,0 +1,52 @@
...
+        $services->set('dashboard.widget.crawler.queuesize')
+            ->class(QueueSizeWidget::class)
+            ->arg('$view', new Reference('dashboard.views.widget'))
+            ->arg('dataProvider', new Reference(QueueSizeDataProvider::class))
+            ->arg('$options', [
+                'title' => 'LLL:EXT:crawler/Resources/Private/Language/locallang.xlf:dashboard.widget.queuesize.widget.title',
+                'icon' => 'content-elements-login'
+            ])
+            ->tag('dashboard.widget', [
+                'identifier' => 'queuesize',
```
brotkrueml, line 42: As the identifier is unique for the whole installation I prefer a prefix like "crawler.queuesize" or something. If not, another extension can perhaps "steal" this identifier.

ext_tables.php
```diff
@@ -5,3 +5,14 @@
     \AOE\Crawler\Utility\BackendUtility::registerInfoModuleFunction();
     \TYPO3\CMS\Core\Utility\ExtensionManagementUtility::allowTableOnStandardPages('tx_crawler_configuration');
 }
+
+
+TYPO3\CMS\Core\Utility\ExtensionManagementUtility::addTypoScriptSetup('
```
brotkrueml, line 10: I suggest a subfolder like "Resources/Private/Templates/Widget". This is the same way the dashboard extension does it. And: If someone decides to use Fluid templates for the Backend module, they are separated from the widgets.

Classes/Widgets/Provider/QueueSizeDataProvider.php
tomasnorre: I know that dashboard will not work in 9 LTS, but will this break the Static Code analysis as it's not supported in 9 LTS with the DI?

CHANGELOG.md
PhilippKuhlmay: I changed this

Configuration/Services.php
PhilippKuhlmay: I changed `reports` to `dashboard`

Configuration/Services.php
PhilippKuhlmay: I changed this to `crawler.queuesize`

ext_tables.php
PhilippKuhlmay: The Fluid template is allready inside the Widget subfolder but I changed the definition in the TypoScript.

Configuration/Services.php
PhilippKuhlmay: Tested without dashboard extension and with TYPO3v9

Configuration/Services.yaml
PhilippKuhlmay: I removed the definition in Services.yaml and put it in Services.php only.

Classes/Widgets/Provider/QueueSizeDataProvider.php
PhilippKuhlmay: I removed the deprecated method and used DI instead

Documentation/Features/Widgets/Index.rst
```diff
@@ -0,0 +1,33 @@
+.. include:: /Includes.txt
+
+=======
+Widgets
+=======
+
+Since 9.1.0
```
brotkrueml, line 7: There is a "versionadded" directive, have a look: https://docs.typo3.org/m/typo3/reference-coreapi/10.4/en-us/_sources/ApiOverview/Mail/Index.rst.txt

I think this should be 9.2.0 (or something) as 9.1.0 is released already.

Resources/Private/Language/locallang.xlf
```diff
@@ -272,7 +272,21 @@
...
+				<source>This preset will give you a dashboard containing the most common Crawler widgets.</source>
+			</trans-unit>
+			<trans-unit id="dashboard.widget.queuesize.title" resname="dashboard.widget.group.description">
+				<source>Queue size</source>
+			</trans-unit>
+			<trans-unit id="dashboard.widget.queuesize.description" resname="dashboard.widget.group.description">
+				<source>Widget for displaying the queue size</source>
+			</trans-unit>
+			<trans-unit id="dashboard.widget.queuesize.widget.title" resname="dashboard.widget.group.description">
+				<source>Size of uncrawled items</source>
```
brotkrueml, line 288: Instead of "size" use better "number"

Classes/Widgets/Provider/QueueSizeDataProvider.php
brotkrueml: > I know that dashboard will not work in 9 LTS, but will this break the Static Code analysis as it's not supported in 9 LTS with the DI?

Don't think so, because the QueueRepository is available. And phpstan doesn't know anything about Symfony DI (I think).

Documentation/Features/Widgets/Index.rst
PhilippKuhlmay: I changed this

PhilippKuhlmay:
This PR can be closed. The widgets moved to a different repository and will be developed there.

PR #621, opened 2020-09-09
Deprecation/count and count all

tomasnorre:
## Description

This deprecates countActive() and getLimitFromItemCountAndOffset() and removes the AbstractRepository as it is not neeeded anymore. 

**I have**


Classes/Domain/Repository/ProcessRepository.php
```diff
@@ -162,20 +156,11 @@ public function removeByProcessId($processId): void
      * Returns the number of active processes.
      *
      * @return int
+     * @deprecated Using ProcessRepository->countActive() is deprecated since 9.1.1 and will be removed in v11.x, please use ProcessRepository->findAllActive->count() instead
```
tomasnorre, line 159: This isn't done with the DeprecatedPublicFunction trait, as the getLimitFromItemCountAndOffset() is public static, and will conflict with the trait.

Classes/Domain/Repository/ProcessRepository.php
```diff
@@ -244,6 +229,7 @@ public function countNotTimeouted(int $ttl): int
 
     /**
      * Get limit clause
+     * @deprecated Using ProcessRepository::getLimitFromItemCountAndOffset() is deprecated since 9.1.1 and will be removed in v11.x, was not used, so will be removed
```
tomasnorre, line 232: This isn't done with the DeprecatedPublicFunction trait, as the getLimitFromItemCountAndOffset() is public static, and will conflict with the trait.

CDRO:
I'm not a big fan of mixing Extbase and DBAL, I'd rather have one way to go as I don't see the benefits of going with Extbase on the performance side of things either.
I see the point of removing the AbstractRepository, though.

tomasnorre:
Thanks for you input @CDRO 

> I'm not a big fan of mixing Extbase and DBAL, I'd rather have one way to go as I don't see the benefits of going with Extbase on the performance side of things either.
> I see the point of removing the AbstractRepository, though.

The reason for me to mix it was that I found it easier to work with objects that the `$this->createQuery()` gives me than the arrays that the DBAL gives. 

I'm totally ok with staying at DBAL if I can have it return objects instead, but for some reason, I haven't figured that out. 
If it's possible I would be happy to stay at DBAL no problem at all.

tomasnorre:
I asked about this in the #TYPO3-CMS Slack that and got the following suggestion, where DBAL and Extbase are combined to get objects.

Not sure how I feel about it though. What do you think @CDRO  @bmack 

```php
public function findAll(): ProcessCollection
    {
        /** @var ProcessCollection $collection */
        $collection = GeneralUtility::makeInstance(ProcessCollection::class);

        $queryBuilder = GeneralUtility::makeInstance(ConnectionPool::class)->getQueryBuilderForTable($this->tableName);

        $queryBuilder
            ->select('*')
            ->from($this->tableName)
            ->execute()->fetchAll();

        /** @var Query $query */
        $query = $this->createQuery();
        $result = $query->statement($queryBuilder)->execute();

        /** @var Process $process */
        foreach ($result as $process) {
            $collection->append($process);
        }

        return $collection;
    }
```

tomasnorre:
I have also been looking into the Doctrine `@ORM` annotations like `@ORM\Column(type="string" name="field-name")` and `@ORM\Entity` but haven't managed to get in working.

CDRO:
I guess you could leave out `execute()` and `fetchAll()` since these will be taken care of by `$query->statement($queryBuilder)->execute()`, but if you want to use a collection and objects of sorts, this is probably quite efficient.

Issue #615, opened 2020-08-06
Display error message in buildQueue command


brotkrueml:
## Feature Request

**Is your feature request related to a problem? Please describe.**
When starting the buildQueue command with a configuration which doesn't work (e.g. a page veto for that specific page is set) the following result is shown: ``Putting 0 entries in queue``. There is not hint, why they are no entries put into the queue where they should some. See also https://github.com/AOEpeople/crawler/issues/516#issuecomment-669759086

**Describe the solution you would like**
The command calls CrawlerController->getPageTreeAndUrls() which returns a "code" (which is HTML). In that code there is probably an error message which should be shown. By now, the command ignores that code. In the described case the codes holds the reason: the page is veto'd, so there are 0 entries in queue.

This would make the configuration error obvious, for now you have to debug to see the error message. I don't know, why it's HTML there (perhaps it's shown in the module?). So the code must be trimmed and striptags'd before output.

tomasnorre:
I'm thinking about how this could be changed. Outputting the HTML doesn't look good in the cli even with strip_tags.
And I'm not sure it gives enough information either. 

I'm considering if the `getPageTreeAndUrls()` and  `drawURLs_addRowsForPage()` should return an object of some type instead.
This would have two advantages:

1. Have HTML and logic separated
2. Having the possibility to format the result fitted for the cli

What do you think about that @brotkrueml ? 

![Screenshot from 2020-08-06 21:57:17](https://user-images.githubusercontent.com/1212481/89576626-d2356a80-d82f-11ea-8975-c2b061ae1487.png)

brotkrueml:
Yes, a small object is pretty. With a renderText() and a renderHtml() method perhaps. And both use a Standalone Fluid template, so the view is separated from the PHP code.

tomasnorre:
Thanks for your input, will let you know when I have a review ready, will not be this weekend though ;)

brotkrueml:
That's fine for me, no hurry.

tomasnorre:
I have a small prove of concept already, would need to check the tests, side effects etc. before I can give you something to review and test, but looks like it works how I imagined it.

tomasnorre:
Doh, I have deleted my branch, so would need to do this again..

PR #636, opened 2020-10-17
[FEATURE] Add progress bar to BuildQueueCommand

### LINKED ISSUE #627: Add progress bar to cli commands
## Feature Request

It would be great to have some progress bar, e.g. https://symfony.com/doc/current/components/console/helpers/progressbar.html for the CLI commands to see and progress. 

**Describe alternatives you've considered**
https://symfony.com/doc/current/components/console/helpers/progressbar.html

carstendietrich:
## Description
Adds a simple Symfony progress bar when using the execute mode to have a visual representation of the crawling progress.

Also removed some dead code from the command since the configuration keys are always returend as array. Fixed a wrong PHPDoc return type in CrawlerController as well.

Resolves #627 partially.

**I have**

### DIFF
--- a/Classes/Command/BuildQueueCommand.php
+++ b/Classes/Command/BuildQueueCommand.php
@@ -26,6 +26,7 @@ use AOE\Crawler\Domain\Model\Reason;
 use AOE\Crawler\Domain\Repository\QueueRepository;
 use AOE\Crawler\Utility\SignalSlotUtility;
 use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Helper\ProgressBar;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
@@ -103,11 +104,11 @@ re-indexing or static publishing from command line.' . chr(10) . chr(10) .
      * Examples:
      *
      * --- Re-cache pages from page 7 and two levels down, executed immediately
-     * $ typo3 crawler:buildQueue 7 defaultConfiguration --depth 2 --mode exec
+     * $ typo3 crawler:buildQueue 7 default --depth 2 --mode exec
      *
      *
      * --- Put entries for re-caching pages from page 7 into queue, 4 every minute.
-     * $ typo3 crawler:buildQueue 7 defaultConfiguration --depth 0 --mode queue --number 4
+     * $ typo3 crawler:buildQueue 7 default --depth 0 --mode queue --number 4
      */
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
@@ -131,15 +132,6 @@ re-indexing or static publishing from command line.' . chr(10) . chr(10) .
         $pageId = MathUtility::forceIntegerInRange($input->getArgument('page'), 0);
 
         $configurationKeys = $this->getConfigurationKeys((string) $input->getArgument('conf'));
-
-        if (! is_array($configurationKeys)) {
-            $configurations = $crawlerController->getUrlsForPageId($pageId);
-            if (is_array($configurations)) {
-                $configurationKeys = array_keys($configurations);
-            } else {
-                $configurationKeys = [];
-            }
-        }
 
         if ($mode === 'queue' || $mode === 'exec') {
             $reason = new Reason();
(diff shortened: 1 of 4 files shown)


carstendietrich:
TODO:
- [ ] Figure out functional tests in DDEV context
- [ ] Update changelog
- [ ] Add tests for progress bar output

tomasnorre:
> TODO:
> 
>     * [ ]  Figure out functional tests in DDEV context

If you go to the `.docker` directory in the repository root, and do `docker-composer up -d` you will get and docker-mysql running. 
After that you can do `composer test:functional` 

Otherwise perhaps a small read here:
https://github.com/AOEpeople/crawler/blob/master/CONTRIBUTING.md#running-tests-without-local-development-environment

carstendietrich:
> > TODO:
> > ```
> > * [ ]  Figure out functional tests in DDEV context
> > ```
> 
> If you go to the `.docker` directory in the repository root, and do `docker-composer up -d` you will get and docker-mysql running.
> After that you can do `composer test:functional`
> 
> Otherwise perhaps a small read here:
> https://github.com/AOEpeople/crawler/blob/master/CONTRIBUTING.md#running-tests-without-local-development-environment

Got the integration tests up and running by reusing the stuff from the composer.json but changing the database values to match DDEV.

Currently, it seems that the progress bar interferes with PHPUnit and fails the test case:
cli

When checking the assertions with XDebug everything looks fine, no clue whats the issue 😭 :
php_debug

Any ideas? So far I only tested commands within unit tests using the Symfony `CommandTester` class..

tomasnorre:
Sorry, I don't have much experience with tests like this perhaps @bmack, @CDRO or @icanhazstring has some input here.

carstendietrich:
Argh, found the issue. Symfony `ProgressBar` outputs to `stderr` which lets PHPUnit think that there is an error. Redirecting the  `stderr` now to `stdout` during the test execution. Tests are now green.

Classes/Command/BuildQueueCommand.php
```diff
@@ -103,11 +104,11 @@ protected function configure(): void
      * Examples:
      *
      * --- Re-cache pages from page 7 and two levels down, executed immediately
-     * $ typo3 crawler:buildQueue 7 defaultConfiguration --depth 2 --mode exec
```
tomasnorre, line 106: Please keep as it was, it is to show that it's the default confugation, not that the configuration is called default in the crawler devbox.

Classes/Command/BuildQueueCommand.php
```diff
@@ -132,15 +133,6 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 
         $configurationKeys = $this->getConfigurationKeys((string) $input->getArgument('conf'));
 
-        if (! is_array($configurationKeys)) {
```
tomasnorre, line 135: Why is this removed??

Tests/Functional/Command/BuildQueueCommandTest.php
```diff
@@ -55,9 +55,9 @@ protected function setUp(): void
      */
     public function buildQueueCommandTest(array $parameters, string $expectedOutput, int $expectedCount): void
     {
-        $commandOutput = '';
+        $commandOutput = [];
```
tomasnorre, line 58: Why change this? the CommandUtility::exec() returns a string

tomasnorre (changes requested):
I haven't tested the functionality yet, just added my comments.

Classes/Command/BuildQueueCommand.php
carstendietrich: `getConfigurationKeys` will always return an array so this code is never reached.

Tests/Functional/Command/BuildQueueCommandTest.php
carstendietrich: Oh seems like a leftover when I fiddled with different commands for executing.. restored the empty string.

Classes/Command/BuildQueueCommand.php
carstendietrich: Ahhh thought it was a bug since the command didn't work in the dev box. good to know, restored it.

tomasnorre (commented):
Thanks, Have a UserGroups as I write, so would need to test this before I can merge. 

But looks good from Code Point of View.

tomasnorre:
Test works as expected too @carstendietrich 
Anything else you want to do before merging?

The pretty output will be handle in #615 also with error handling.

carstendietrich:
> Test works as expected too @carstendietrich
> Anything else you want to do before merging?
> 
> The pretty output will be handle in #615 also with error handling.

Nope, guess I'm fine. Let's merge this one.

tomasnorre (approved):

tomasnorre:
It should actually have been a 9.2.0 version as it's a feature, but just say that after click merge. No big deal. It's at least not a breaking change.

PR #644, opened 2020-10-29
[DOC] Update Documentation regarding Cache WarmUp and UserGroups

### LINKED ISSUE #639: Improve documentation about crawler configuration / cache warmup
## Feature Request

**Is your feature request related to a problem? Please describe.**
I have access restricted pages and I did not understand directly how the caching in TYPO3 works. As a result I created just a single crawler configuration although my user can have multiple `usergroup` configuration. 

**Describe the solution you would like**
A hint in the documentation that a proper cache warmup might need multiple crawler configurations

**Describe alternatives you've considered**
In chapter [Cache warm up](https://docs.typo3.org/p/aoepeople/crawler/9.1/en-us/UseCases/CacheWarmup/Index.html) I would like to see the hint (maybe really in such a box with an exclamation mark) with something like 

"On access restricted pages you need a crawler configuration for every usergroup combination that a user could have who shall visit the page. An example: Your page is restricted with usergroup _Member_ and your users could only have this usergroup but also a usergroup _Premium Member_. In this case you have to create two crawler configurations: Configuration 1 contains as usergroup only _Member_, configuration 2 contains as usergroups _Member_ and _Premium Member_."

The wording shall not be seen as finalized...

rintisch:
## Description

Improve documentation, fix issue described in #639 

**I have**

(sorry, I didn't know that I should add a description in CHANGELOG.md. Furthermore was it the first time I added somethin in reST -  so I hope it's okay how it's made)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Added
 * Progress bar to the `crawler:buildQueue` command output when using with `--mode exec`
+* Improve documentation about proper crawler configuration for user with _Website Usergroups_
 
 ### Deprecated
 
--- a/Documentation/Configuration/ConfigurationRecords/Index.rst
+++ b/Documentation/Configuration/ConfigurationRecords/Index.rst
@@ -60,7 +60,7 @@ Processing instruction parameters
 
 Crawl with FE user groups
    User groups to set for the request. See also:
-   :ref:`paramSets.[key].userGroups <crawler-tsconfig-paramSets-key-userGroups>`
+   :ref:`paramSets.[key].userGroups <crawler-tsconfig-paramSets-key-userGroups>` and the hint in :ref:`create-crawler-configuration`
 
 Access
 ------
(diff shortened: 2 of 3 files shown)


tomasnorre (changes requested):
The rendering isn't correct I think? Think you need to wrap it to "escape" the spacing in between. 

Screenshot from 2020-10-29 13:54:03

rintisch:
Okay, I will give it another try. Can I check it somehow myself?

tomasnorre:
Yes. 

In the Crawler root folder 
```
docker run --rm t3docs/render-documentation:v2.6.1
source <(docker run --rm t3docs/render-documentation:v2.6.1 show-shell-commands)
dockrun_t3rd makehtml                                                           
firefox Documentation-GENERATED-temp/Result/project/0.0.0/Index.html
```

rintisch:
@tomasnorre I scrolled through the TYPO3 documentation and realized that proper nouns are nowhere set as italic. So I removed this formatting.

PR #663, opened 2020-11-29
[BUGFIX] Assign port as string to $_SERVER['port'] in cli/bootstrap.php

### LINKED ISSUE #662: Bug in cli command 'crawler:processQueue' related to 'trustedHostsPattern'
## Bug Report

**Current Behavior**
When using cli command 'crawler:processQueue' the following exception is thrown

Uncaught TYPO3 Exception #1396795884: The current host header value does not match the configured trusted hosts pattern! Check the pattern defined in $GLOBALS['TYPO3_CONF_VARS']['SYS']['trustedHostsPattern'] and adapt it, if you want to allow the current host header 'mydomain.de:8077' for your installation.
thrown in file /mypath/to/typo3root/typo3/sysext/core/Classes/Utility/GeneralUtility.php
in line 2722

It's only affected if no standard port is used

**Expected behavior/output**
No exception is throw.

**Environment**
- Crawler version(s): 9.1.4
- TYPO3 version(s): 10.4.10

**Possible Solution**
In file 'cli/bootstrap.php' line 73 just convert '$_SERVER['SERVER_PORT']' to string as in TYPO3 port is only comparing this value to a string.
`$_SERVER['SERVER_PORT'] = (string)$urlParts['port'];`

**Additional context**

See: `\TYPO3\CMS\Core\Utility\GeneralUtility::hostHeaderValueMatchesTrustedHostsPattern()`:
```php
            $parsedHostValue = parse_url('http://' . $hostHeaderValue);
            if (isset($parsedHostValue['port'])) {
                $hostMatch = (strtolower($parsedHostValue['host']) === strtolower($_SERVER['SERVER_NAME']) && (string)$parsedHostValue['port'] === $_SERVER['SERVER_PORT']);
            } else {
                $hostMatch = (strtolower($hostHeaderValue) === strtolower($_SERVER['SERVER_NAME']) && $defaultPort === $_SERVER['SERVER_PORT']);
            }

´´´

### PR DESCRIPTION
**Pull request by gelbehexe**
## Description

Resolves #662: Bug in cli command 'crawler:processQueue' related to 'trustedHostsPattern'

### DIFF
--- a/cli/bootstrap.php
+++ b/cli/bootstrap.php
@@ -70,7 +70,7 @@ $_SERVER['REQUEST_METHOD'] = 'GET';
     // Define a port if used in the URL:
 if (isset($urlParts['port'])) {
     $_SERVER['HTTP_HOST'] .= ':' . $urlParts['port'];
-    $_SERVER['SERVER_PORT'] = $urlParts['port'];
+    $_SERVER['SERVER_PORT'] = (string)$urlParts['port'];
 }
 
     // Define HTTPS disposal:

### CODE REVIEW

**Comment by tomasnorre** [comment]
Can you please write how to reproduce this? 

Do you set the baseUrl in crawler configuration with port? Where do you configure your URL with port? 
As long as I cannot reproduce, I cannot test and merge. 

Looks ok to me, but want to verify and at best also add a test for it.

**Comment by gelbehexe** [comment]
Ich used enviroment variable "BASE_URL" which looks like BASE_URL="http://mydomain.de:8077/". It is set by .env in a docker compose container. This variable is used in site configuration like this:
```yaml
baseVariants:
  -
    base: '%env(BASE_URL)%'
    condition: 'getenv("BASE_URL")'
```
I xdebugged your code in cli mode until I came to the point where './typo3conf/ext/crawler/cli/bootstrap.php' is executed as shell command. I copied this command line before it is executed  and xdebugged this command. So I'm absolutly sure about the reason why comparing the port part in GeneralUtility fails: `parse_url()` result array with key 'port' contains always an integer value und is compared with a string by operator `===` which fails.

tomasnorre:
Thanks I think I will be able to reproduce with the additional information and provide a test for it.

tomasnorre:
It sounds like that you are using direct request in the Crawler settings, can you please confirm that? 
And if so could you also let me know if it's only a problem with direct request or also with direct request disabled? 

Thanks

tomasnorre:
I cannot reproduce this problem, but I can also not see a problem with your code change, so will merge this and see if I can find a way to have test written.

Issue #649, opened 2020-11-11
[BUG] TypeError: Argument 1 passed to CrawlerController::getUrlsForPageId() must be of the type int, null given


PhilippKuhlmay:
## Bug Report

**Current Behavior**
When I create a scheduler task with the task crawler:buildQueue and set the option `page` to 0, the task throws an error.
<img width="1143" alt="Bildschirmfoto 2020-11-11 um 09 18 39" src="https://user-images.githubusercontent.com/57705869/98787203-7b779800-23ff-11eb-8fe4-59daf95f3b84.png"> 
When I set the page to 1, it works as expected.

**Expected behavior/output**
I expected the task to be executed (_or not be executed because there is no config on pageId 0_) without an error. 

**Environment**
- Crawler version(s): 9.1.0
- TYPO3 version(s): 10.4.9
- Is your TYPO3 installation set up with Composer (Composer Mode): yes

**Additional context**
My pagetree:
<img width="206" alt="Bildschirmfoto 2020-11-11 um 09 19 43" src="https://user-images.githubusercontent.com/57705869/98787270-94804900-23ff-11eb-8c31-e3320c2f3a11.png">

My scheduler task config:
<img width="752" alt="Bildschirmfoto 2020-11-11 um 09 19 03" src="https://user-images.githubusercontent.com/57705869/98787299-9cd88400-23ff-11eb-9ca7-481562544bc7.png">

The crawler config is on page 1 - Startseite.

tomasnorre:
Please try it out with 1 as start page. 0 I'm not sure it works.

PR #675, opened 2020-12-06
[TASK] Add error handling when page 0 is used for buildQueue

tomasnorre:
## Description

Adds Error information when page 0 is used for crawler:buildQueue

Resolves: #649 

**I have**


tomasnorre:
@PhilippKuhlmay Can you please check if this matches your expectations.
I didn't find a way to add it as a check on save already. Would have been better though.

PR #721, opened 2021-02-05
[CLEANUP] Remove extension png icon

brotkrueml:
## Description

The svg extension icon is preferred, so no need to provide
a png icon additionally.

See: https://docs.typo3.org/m/typo3/reference-coreapi/9.5/en-us/ExtensionArchitecture/FilesAndLocations/Index.html#resources-public-icons-extension-svg


brotkrueml:
@tomasnorre Hmm, seems the error is not on my side: The PR branch could not be required by composer?

brotkrueml:
> @tomasnorre Hmm, seems the error is not on my side: The PR branch could not be required by composer?

Okay, with PR #723 the problem should be fixed :-)

PR #725, opened 2021-02-22
Explain _RECURSIVE flag in documentation

kraemer-igroup:
Hi there, I recently searched for an option to crawl multiple news folders at once. Within the code I found the _RECURSIVE flag which is not explained in the documentation yet (https://docs.typo3.org/p/aoepeople/crawler/9.2/en-us/Configuration/PageTsconfigReference(txCrawlercrawlercfg)/Index.html).


tomasnorre:
Thanks for you PR.

Issue #729, opened 2021-03-18
Indexing wasnt forced anymore since middleware order has changed


rengaw83:
Indexing was'nt forced anymore.

TYPO3 9.5.25
Crawler: 9.2.2

The urls are crawled, but the indexing wasn't forced.

In have debugged a lot, and in my case the problem is the current middleware registration order.
In #642 the order was changed, to fix issues with StaticFileCache extension.
See https://github.com/AOEpeople/crawler/commit/0f7cb6abac4d8d6e228b0385665dc02f17cdfdcb#diff-38337bad08776b0fe94f73f1cf471f8d184ec3f9d2e9254f7c2a275b83501e34

The middleware order (without staticfilecache installed) now looks like this:
```
    typo3/cms-frontend/timetracker
    typo3/cms-core/normalized-params-attribute
    typo3/cms-frontend/preprocessing
    typo3/cms-frontend/eid
    typo3/cms-frontend/tsfe
    typo3/cms-frontend/authentication
    typo3/cms-frontend/site
    typo3/cms-frontend/preview-simulator
    aoe/crawler/authentication
    typo3/cms-frontend/base-redirect-resolver
    typo3/cms-frontend/static-route-resolver
    typo3/cms-frontend/page-resolver
    typo3/cms-frontend/maintenance-mode
    typo3/cms-frontend/page-argument-validator
    typo3/cms-frontend/prepare-tsfe-rendering
    typo3/cms-frontend/shortcut-and-mountpoint-redirect
    typo3/cms-frontend/content-length-headers
    aoe/crawler/initialization
    typo3/cms-frontend/output-compression
```

The `aoe/crawler/initialization` middleware is loaded after the `typo3/cms-frontend/prepare-tsfe-rendering` now. 
The crawler can't fill `$GLOBALS['TSFE']->applicationData['tx_crawler']` before the TSFE checks the `headerNoCache` and so indexed_search  cant disable cache output. so no page is generated and no content can be indexed.

Here typo3 9 checks if the page should be generated or the cache will be used
https://github.com/TYPO3/TYPO3.CMS/blob/9.5/typo3/sysext/frontend/Classes/Controller/TypoScriptFrontendController.php#L2498-L2519

Here indexed search checks if the crawler is enabled and a indexing is in progress
https://github.com/TYPO3/TYPO3.CMS/blob/9.5/typo3/sysext/indexed_search/Classes/Hook/TypoScriptFrontendHook.php#L25-L40

[...]

PR #735, opened 2021-04-13
[BUGFIX] Add new middleware for content finish

tomasnorre:
This middleware splits the content from the Crawler Initialization Middleware
to ensure that the content is written at the end.

This ensures that middleware that expects response-object gets the content
for the renders and the crawler queue gets the correct request status.

Resolves: #729

**I have**


tomasnorre:
@CDRO I could not assign you as reviewer, would you mind looking at the PR?

CDRO:
@tomasnorre Looks good to me!

tomasnorre:
Hi @lochmueller,

I was just looking into stalled PRs, could this perhaps be the solution that would also solve the problem the crawler is causing for the Staticfilecache? 

I would be happy to hear you feedback.

lochmueller:
Hey @tomasnorre 
This move "the problem" just up in the middleware stack, so StaticFileCache would work. But it still detroy the middleware stack, because a "new response" is created. I send a new pull request incl. a better solution in a few minutes...
Regards,
Tim

tomasnorre:
Closed with #1018

PR #747, opened 2021-05-23
[REFACTOR] Add cli methods to ProcessCommand and deprecated them in CrawlerController

tomasnorre:
## Description

This PR Add the CLI_ functions from the CrawlerController to ProcessCommand, there are only used in there, so having the as private functions in there instead of a Service was more logical to me, but still not 100% sure if I like the service approach more to have the ProcessCommand smaller.

**I have**

- [X] Added tests for the new code

### DIFF
--- a/Classes/Command/ProcessQueueCommand.php
+++ b/Classes/Command/ProcessQueueCommand.php
@@ -326,18 +326,17 @@ class ProcessQueueCommand extends Command
 
     private function getCrawlerController(): CrawlerController
     {
-        $objectManager = GeneralUtility::makeInstance(ObjectManager::class);
-        return $this->crawlerController ?? $objectManager->get(CrawlerController::class);
+        return $this->crawlerController ?? GeneralUtility::makeInstance(CrawlerController::class);
     }
 
     private function getProcessRepository(): ProcessRepository
     {
-        return $this->processRepository ?? new ProcessRepository();
+        return $this->processRepository ?? GeneralUtility::makeInstance(ProcessRepository::class);
     }
 
     private function getQueueRepository(): QueueRepository
     {
-        return $this->queueRepository ?? new QueueRepository();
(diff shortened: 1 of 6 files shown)


Classes/Command/ProcessQueueCommand.php
```diff
@@ -147,4 +193,155 @@ protected function configure(): void
...
+
+    private function getCrawlerController(): CrawlerController
+    {
+        $objectManager = GeneralUtility::makeInstance(ObjectManager::class);
+        return $this->crawlerController ?? $objectManager->get(CrawlerController::class);
+    }
+
+    private function getProcessRepository(): ProcessRepository
+    {
+        return $this->processRepository ?? new ProcessRepository();
```
CDRO, line 335: Wouldn't you like to use `GeneralUtility::makeInstance` here as well?

Classes/Command/ProcessQueueCommand.php
```diff
@@ -147,4 +193,155 @@ protected function configure(): void
...
+    }
+
+    private function getQueueRepository(): QueueRepository
+    {
+        return $this->queueRepository ?? new QueueRepository();
```
CDRO, line 340: See above about using `GeneralUtility::makeInstance`

Classes/Command/ProcessQueueCommand.php
```diff
@@ -147,4 +193,155 @@ protected function configure(): void
...
+
+    private function getCrawler(): Crawler
+    {
+        return $this->crawler ?? new Crawler();
+    }
+
+    private function getCrawlerController(): CrawlerController
+    {
+        $objectManager = GeneralUtility::makeInstance(ObjectManager::class);
+        return $this->crawlerController ?? $objectManager->get(CrawlerController::class);
```
CDRO, line 330: I would mark this as TODO for the next version so you just use `GeneralUtility::makeInstance` since it takes care of DI and `ObjectManager->get` should soon be deprecated.

PR #748, opened 2021-05-23
[REFACTOR] Swicht to FetchAssociative() instead of fetch() as deprecated

tomasnorre:
## Description

**I have**


tomasnorre:
This will be rejected first, as not supported as long as we have TYPO3 v9 support. I have created an issue for this instead, to not forget about it. 

https://github.com/AOEpeople/crawler/issues/749

PR #754, opened 2021-06-15
[FEATURE] Display error messages in buildQeueu command

### LINKED ISSUE #615: Display error message in buildQueue command

tomasnorre:
## Description

Resolves #615 

**I have**


tomasnorre:
Hi @brotkrueml 

Would you mind reviewing and test this? It's my draft suggestion for your feature request #615 

Thanks

Classes/Controller/CrawlerController.php
```diff
@@ -1288,10 +1292,11 @@ public function getPageTreeAndUrls(
                 $mountTree->getTree($mountpage[0]['mount_pid'], $depth);
 
                 foreach ($mountTree->tree as $mountData) {
-                    $code .= $this->drawURLs_addRowsForPage(
+                    $queueRows = array_merge($queueRows, $this->drawURLs_addRowsForPage(
```
brotkrueml, line 1295: Why `array_merge()` and not `$queueRows[] = $this->drawURLs_addRowsForPage(...);`?

Classes/Controller/CrawlerController.php
```diff
@@ -1303,13 +1308,14 @@ public function getPageTreeAndUrls(
                 }
             }
 
-            $code .= $this->drawURLs_addRowsForPage(
+            $queueRows = array_merge($queueRows, $this->drawURLs_addRowsForPage(
```
brotkrueml, line 1311: Same here

Classes/Controller/CrawlerController.php
```diff
@@ -1385,47 +1395,40 @@ public function drawURLs_addRowsForPage(array $pageRow, string $pageTitle): stri
...
                     // Options
-                    $optionValues = '';
+                    $queueRowOptionCollection = [];
                     if ($confArray['subCfg']['userGroups']) {
-                        $optionValues .= 'User Groups: ' . $confArray['subCfg']['userGroups'] . '<br/>';
+                        $queueRowOptionCollection[] = 'User Groups: ' . $confArray['subCfg']['userGroups'];
                     }
                     if ($confArray['subCfg']['procInstrFilter']) {
-                        $optionValues .= 'ProcInstr: ' . $confArray['subCfg']['procInstrFilter'] . '<br/>';
+                        $queueRowOptionCollection[] = 'ProcInstr: ' . $confArray['subCfg']['procInstrFilter'];
```
brotkrueml, line 1403: Same here

Classes/Controller/CrawlerController.php
```diff
@@ -1385,47 +1395,40 @@ public function drawURLs_addRowsForPage(array $pageRow, string $pageTitle): stri
...
-                        <tr>
-                            ' . $titleClm . '
-                            <td>' . htmlspecialchars($confKey) . '</td>
-                            <td>' . nl2br(htmlspecialchars(rawurldecode(trim(str_replace('&', chr(10) . '&', GeneralUtility::implodeArrayForUrl('', $confArray['paramParsed'])))))) . '</td>
-                            <td>' . $paramExpanded . '</td>
-                            <td nowrap="nowrap">' . $urlList . '</td>
-                            <td nowrap="nowrap">' . $optionValues . '</td>
-                            <td nowrap="nowrap">' . DebugUtility::viewArray($confArray['subCfg']['procInstrParams.']) . '</td>
-                        </tr>';
+                    $parameterConfig = nl2br(htmlspecialchars(rawurldecode(trim(str_replace('&', chr(10) . '&', GeneralUtility::implodeArrayForUrl('', $confArray['paramParsed']))))));
```
brotkrueml, line 1406: Same here

Classes/Controller/CrawlerController.php
```diff
@@ -1385,47 +1395,40 @@ public function drawURLs_addRowsForPage(array $pageRow, string $pageTitle): stri
...
+                    $queueRow->setValuesExpanded($paramExpanded);
+                    $queueRow->setConfigurationKey($confKey);
+                    $queueRow->setUrls($urlList);
+                    $queueRow->setOptions($queueRowOptionCollection);
+                    $queueRow->setParameters(DebugUtility::viewArray($confArray['subCfg']['procInstrParams.']));
```
brotkrueml, line 1411: Same here

Classes/Value/QueueRow.php
```diff
@@ -0,0 +1,98 @@
...
+
+    /** @var string */
+    public $parameters;
+
+    public function __construct(string $title = '')
+    {
+        $this->pageTitle = $title;
+    }
+
+    public function setPageTitleHTML(string $pageTitleHTML): void
```
brotkrueml, line 59: Why setters if you declare the class as internal and the properties as public?

tomasnorre: Because my IDE doesn't complain about types if doing false assignments, which the example is. 

```php
$queueRow->message = [];
```

Classes/Controller/CrawlerController.php
```diff
@@ -1385,47 +1395,40 @@ public function drawURLs_addRowsForPage(array $pageRow, string $pageTitle): stri
                     $paramExpanded .= 'Comb: ' . implode('*', $calcAccu) . '=' . $calcRes;
 
                     // Options
-                    $optionValues = '';
+                    $queueRowOptionCollection = [];
                     if ($confArray['subCfg']['userGroups']) {
-                        $optionValues .= 'User Groups: ' . $confArray['subCfg']['userGroups'] . '<br/>';
+                        $queueRowOptionCollection[] = 'User Groups: ' . $confArray['subCfg']['userGroups'];
```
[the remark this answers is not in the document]
tomasnorre: Not always. I'll consider a better option, don't want the output with empty user groups if none. 

```
User Groups: 
```

tomasnorre:
Did you have a chance to test it @brotkrueml ?

brotkrueml:
> Did you have a chance to test it @brotkrueml ?

Yes, will do in the next days. Don't have page vetos anymore, so have to create one.

tomasnorre:
I think the output alone will give you an idea if it works or not, as Folders etc. will be echoed as warning with Doktype not supported.

brotkrueml:
Okay, it is very chatty ;-).

----

Now a URL with MP is used:

    Processing
    https://website.ddev.site:8443/en/?MP= () => 
    OK: 

This shouldn't be an issue, as the correct canonical is used (without MP), just a little bit unaesthetic.

----

I use the following configuration (3 is the start page):

    tx_crawler.crawlerCfg.paramSets {
        deployment = &L=[0-3]
        deployment {
            pidsOnly = 3
        }
    }

With this configuration:

    crawler:buildQueue 3 deployment --depth 99 --mode exec

So, it should only generate the start pages for four languages. But I get information from all other pages (Page-x are my placeholder for the real name):

    Page-1:  (Because page is hidden)
    Page-2:  (Because page is hidden)
    Page-3:  (Because doktype is not allowed)

This is not very helpful as only one page in different languages should be generated.

Yes, I use `depth`, but the `pidsOnly` should overrule this IMHO.

----

Running the command without `depth`:

    ddev t3cmd crawler:buildQueue 3 deployment --mode exec                                                                                                                                
    
    Executing 4 requests right away:
    [20.06.21 17:19] https://website.ddev.site:8443/en/?MP= (URL already existed)<br>[20.06.21 17:19] https://website.ddev.site:8443/de/?MP= (URL already existed)<br>[20.06.21 17:19] https://website.ddev.site:8443/pl/?MP= (URL already existed)<br>[20.06.21 17:19] https://website.ddev.site:8443/tr/?MP= (URL already existed)

omits the detailled information from above for other pages. The `<br>` tag should be converted to a new line on console.

----

When I run the command for a non-existing page:

    crawler:buildQueue 9999999 deployment --depth 99 --mode exec

The following output is given:

    Executing 0 requests right away:
    Processing

        0 [->--------------------------]

Perhaps it is better to give an error, that the page does not exist.

----

I am getting many empty lines when calling a `buildQueue` command with `depth`. Perhaps these empty lines come from "successful" pages without any output. I think, they should be avoided.

----

What does the message `Doktype was excluded by "0"` mean? This is given on a page with doktype "Backend user section".

----

Because it is very chatty, I think it would be a possibility to hide the output behind an option, perhaps `--verbose` or `--debug`.

----

At at the end: I like it, it is possible now to get the information needed for a given page, why it is not crawled. Well done :-)

tomasnorre:
Thanks for your feedback @brotkrueml. I'll have a look at it.

Classes/Controller/CrawlerController.php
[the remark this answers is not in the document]
tomasnorre: Sorry for the really long response time @brotkrueml 

What do you think about this solution instead? It's a modified version of your suggestion. 
```php
$queueRowOptionCollection = [];
$queueRowOptionCollection[] = $confArray['subCfg']['userGroups'] ? 'User Groups: ' . $confArray['subCfg']['userGroups'] : '';
$queueRowOptionCollection[] = $confArray['subCfg']['procInstrFilter'] ? 'ProcInstr: ' . $confArray['subCfg']['procInstrFilter'] : '';

// Remove empty array entries;
$queueRowOptionCollection = array_filter($queueRowOptionCollection);
```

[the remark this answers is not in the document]
brotkrueml: Hmm, I would prefer something like
```
if ($confArray['subCfg']['userGroups'] ?? false) {
    $queueRowOptionCollection[] = 'User Groups: ' . $confArray['subCfg']['userGroups'];
}
```
to avoid adding something to an array and then removing it later. It reads better (had to think some seconds about ```array_filter``` without callback ;-) - haven't known the callback is optional).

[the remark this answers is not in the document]
tomasnorre: Thanks, for the feedback, then I got you wrong. Like you suggestion better too.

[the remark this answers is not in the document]
brotkrueml: Wanted to give you just the direction of non-existing keys.

tomasnorre:
Hi @brotkrueml 

As the main feature is now "done", I would suggest that I merge this and do some fixes later.
The reason is that this has breaking changes, and would be a good fit to include in the 11.0.0 release of the Crawler as there are quite some breaking changes too. 

I would then take your comments https://github.com/tomasnorre/crawler/pull/754#issuecomment-864586872 and create an issue with it, so that I can address this later. 

As I read them, it's improvements and not "bugs"

What do you think about this aproach ?

PR #835, opened 2021-10-29
[BUGFIX] Ensure that Crawler does  not break Frontend plugins

tomasnorre:
## Description

**I have**


tomasnorre:
Would you mind review and test this @brotkrueml ?

Classes/Middleware/FrontendUserAuthenticator.php
```diff
@@ -48,13 +48,15 @@ class FrontendUserAuthenticator implements MiddlewareInterface
 
     /**
      * @var QueueRepository
+     * This is not used by the Crawler.
```
brotkrueml, line 51: That's why middlewares (and other classes as well) should be declared final ;-)

I would remove it, as it is null now and by that breaking. A possible user who extends this middleware (which is unlikely) cannot rely on it anymore.

Classes/Middleware/FrontendUserAuthenticator.php
```diff
@@ -48,13 +48,15 @@ class FrontendUserAuthenticator implements MiddlewareInterface
...
+     * This is only kept, to not have a breaking change is this bugfix release.
+     * Will be removed in v12.0
      */
     protected $queueRepository;
 
     public function __construct(?Context $context = null)
```
brotkrueml, line 57: I would suggest to inject the QueryBuilder via DI:

Services.yaml:

    queryBuilder.tx_crawler_queue:
      class: 'TYPO3\CMS\Core\Database\Query\QueryBuilder'
      factory: ['@TYPO3\CMS\Core\Database\ConnectionPool', 'getQueryBuilderForTable']
      arguments: ['tx_crawler_queue']

    AOE\Crawler\Middleware\FrontendUserAuthenticator:
      arguments:
        $queryBuilder: '@queryBuilder.tx_crawler_queue' 

Constructor:

    public function __construct(QueryBuilder $queryBuilder, ?Context $context = null)
    {
        $this->queryBuilder = $queryBuilder;
        $this->context = $context ?? GeneralUtility::makeInstance(Context::class);
    }

This avoids GeneralUtility::makeInstance() call and makes possible unit tests easier, as a mock of the query builder can be injected if needed. Also you can reuse the `queryBuilder.tx_crawler_queue` reference later for other classes.

Classes/Middleware/FrontendUserAuthenticator.php
```diff
@@ -121,4 +123,18 @@ private function getFrontendUser(string $grList, ServerRequestInterface $request
         $frontendUser->user['uid'] = PHP_INT_MAX;
         return $frontendUser;
     }
+
+    private function findByQueueId(string $queueId): array
+    {
+        $queryBuilder = GeneralUtility::makeInstance(ConnectionPool::class)->getQueryBuilderForTable(QueueRepository::TABLE_NAME);
```
brotkrueml, line 129: As already mentioned above, using DI you can omit GeneralUtility::makeInstance()

Classes/Middleware/FrontendUserAuthenticator.php
```diff
@@ -121,4 +123,18 @@ private function getFrontendUser(string $grList, ServerRequestInterface $request
...
+        $queueRec = $queryBuilder
```
brotkrueml, line 130: Use $this->queryBuilder() when using the suggested DI variant.

Classes/Middleware/FrontendUserAuthenticator.php
```diff
@@ -121,4 +123,18 @@ private function getFrontendUser(string $grList, ServerRequestInterface $request
...
+            ->select('*')
+            ->from(QueueRepository::TABLE_NAME)
+            ->where(
+                $queryBuilder->expr()->eq('qid', $queryBuilder->createNamedParameter($queueId))
```
brotkrueml, line 134: Use $this->queryBuilder() when using the suggested DI variant.

Classes/Middleware/FrontendUserAuthenticator.php
```diff
@@ -121,4 +123,18 @@ private function getFrontendUser(string $grList, ServerRequestInterface $request
...
+            )
+            ->execute()
+            ->fetchAssociative();
```
brotkrueml, line 137: The legacy (classic) installation of TYPO3 v10.4 is delivered with DBAL 2.10, which doesn't know about `fetchAssociative()`:

https://github.com/TYPO3/typo3/blob/10.4/composer.lock#L220

So, the (now) deprecated `fetch()` method should be used to be compatible with legacy installations.

Classes/Middleware/FrontendUserAuthenticator.php
```diff
@@ -48,13 +48,15 @@ class FrontendUserAuthenticator implements MiddlewareInterface
 
     /**
      * @var QueueRepository
+     * This is not used by the Crawler.
+     * This is only kept, to not have a breaking change is this bugfix release.
```
brotkrueml, line 52: typo: is = in

brotkrueml:
> Would you mind reviewing and test this @brotkrueml ?

Tested in TYPO4 v11.5.1 and the exception has vanished.

Classes/Middleware/FrontendUserAuthenticator.php
tomasnorre: With legacy installations you mean non-composer based? Because with composer, it's fetch newer dependencies.

brotkrueml: Yup, it the new wording:

https://docs.typo3.org/m/typo3/tutorial-getting-started/master/en-us/Installation/LegacyInstallation.html

tomasnorre: I'll revert that then.

brotkrueml: Yup, I assume, that TYPO3 v12 will be compatible with dbal 3.x, and as TYPO3 v11 is compatible with dbal 2.13, once compatibility with TYPO3 v10 is removed, the `fetchAssociative()` method can be used then.

tomasnorre:
Thanks for the review @brotkrueml. 

The input with DI is valid, but I don't want to introduce a Breaking Change, so I have added it for my reminder for 12.0.0 release. 

I'll in general make all DIs for the 12 release.

Thanks again for your help and input.

PR #839, opened 2021-10-29
[TASK] Normalize header comment

brotkrueml:
This is done via a PHP CS Fixer rule in ecs.php. As this changes almost every
PHP file conflicts to other pull requests are imminent.

See: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/doc/rules/comment/header_comment.rst

### DIFF
--- a/Classes/Api/CrawlerApi.php
+++ b/Classes/Api/CrawlerApi.php
@@ -2,31 +2,22 @@
 
 declare(strict_types=1);
 
-namespace AOE\Crawler\Api;
+/*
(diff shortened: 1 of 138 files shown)


tomasnorre:
Unfortunately yes. I correct them as I see fit. 
I once in a while go through the files with git blame and correct the ones with only changes from the ones approved. 

I'm not 100% sure in which way I'm allowed to change the License. But will look more into this.

brotkrueml:
Okay, then I will close this PR for now, so it is not hanging around.

PR #841, opened 2021-10-30
[BUGFIX] PHP 8.0 compatibility problems. Undefined indexes resolved

tomasnorre:
There are still something with the devbox that isn't working for PHP 8.0 so I'll pause this for now.


tomasnorre:
Rebase when #835 has been merged

Issue #837, opened 2021-10-29
Support StaticFileCache and do not destroy the middleware stack


lochmueller:
## Bug Report

**Current Behavior**
The current problem is, that EXT:crawler destroy the middleware stack, by calling the handler and ignore the response in the middleware:
https://github.com/tomasnorre/crawler/blob/main/Classes/Middleware/CrawlerInitialization.php#L69
So StaticFileCache but also other middlewares in the middleware stack got an useless response object.

The reason for that is, because there are some serialized meta information in the Response, that are use by the crawler itself:
https://github.com/tomasnorre/crawler/blob/main/Classes/Middleware/CrawlerInitialization.php#L80

**Expected behavior/output**
A valid response object that is generated by the middleware stack and not a fresh response object.

**Steps to reproduce**
Using EXT:crawler and EXT:staticfilecache (without the "NoCrawler" rule, because that currently check if the request is a crawler request and do not cache then)

**Environment**
- Crawler version(s): dev-master
- TYPO3 version(s): 10 & 11
- Is your TYPO3 installation set up with Composer (Composer Mode): yes

**Possible Solution**
Write the crawler meta information in some kind of header like "X-Crawler" of the original response object. So the middleware stack is not compromised be the crawler middleware. You can evaluate the Header instead of the content in the the crawling strategy: https://github.com/tomasnorre/crawler/blob/main/Classes/CrawlStrategy/GuzzleExecutionStrategy.php#L54

Then StaticFileCache get a response object with the right content and I can drop the "Disable Static File Cache with Crawler calls"-rule in EXT:staticfilecache.

Any addional ideas?!

Regards,
Tim

tomasnorre:
I have tried to setup staticfilecache together with the crawler, but I don't find where this config "noCrawler" can be set. I don't find anything in the documentation. Can you please point me to this @lochmueller ?

lochmueller:
Hey @tomasnorre 
you could comment out this lines:
https://github.com/lochmueller/staticfilecache/blob/master/Classes/Cache/Rule/NoCrawlerCall.php#L21
https://github.com/lochmueller/staticfilecache/blob/master/Classes/Cache/Rule/NoCrawlerCall.php#L22

Then the cache rule do not disable the cache on crawler calls.
There is no configuration option, because it don't work with the EXT:crawler currently.

If you disable the rule, SFC will cache your serialized meta data information in the StaticFileCache. If the middleware return the real content as described in this issue here, the HTML content should be cached.

[...]

PR #842, opened 2021-10-30
Related #837 - use header instead of content for meta information

lochmueller:
## Description

My idea to solve this problem: Please double check this. Didn't test it in detail.
https://github.com/tomasnorre/crawler/issues/837


tomasnorre:
I have fixed the tests this morning, could you please rebase your `fix-middleware-stack` branch based on my branch with the same name. I didn't manage to push to have the changes within the same PR.

lochmueller:
Done. I fast-forward merge your changes into this PR

tomasnorre:
Thanks. I'll have this tested tomorrow. But from the code it looks gode.

tomasnorre:
I have tested this now, and everything works like I expected. StaticFileCache with NoCrawl like

```php
class NoCrawlerCall extends AbstractRule
{
    public function checkRule(ServerRequestInterface $request, array &$explanation, bool &$skipProcessing): void
    {
        $crawlerInformation = $request->getHeaderLine('X-T3CRAWLER') ?? null;
        if (!empty($crawlerInformation)) {
            // @todo find another solution to cache content via Crawler (avoid caching the status JSON)
            // @see https://github.com/lochmueller/staticfilecache/issues/260
            //$explanation[__CLASS__] = 'It is a Crawler configuration call, that should not cached by SFC';
            //$skipProcessing = true;
        }
    }
}
```

Let the crawler also fill the StaticFileCache. 

My suggestion would be, that your next release conflicts with `tomasnorre/crawler` 11.0.0 or lower, to ensure that people are updating to the compatible crawler version. 

Thanks a lot for your help with this. Really appreciated.

tomasnorre (approved):

Issue #865, opened 2021-12-21
“20.12.21 14:51” could not be parsed by \DateTime constructor: DateTime::__construct()


BokuNoMaxi:
## Bug Report
If the DateFormat in the LocalConfiguration is set to a non accepted DateTime Format,  the ShowLog.html throws “20.12.21 14:51” could not be parsed by \DateTime constructor: DateTime::__construct()
![Error](https://user-images.githubusercontent.com/8913243/146895551-da6ec09b-bb06-47c8-b034-e83e0d6469b9.png)

**Current Behavior**
Opening Info from the Sidebar and select Site-Crawler the Exception: “20.12.21 14:51” could not be parsed by \DateTime constructor: DateTime::__construct() gets thrown

**Expected behavior/output**
All Information regarding the SiteCrawler gets displayed

**Steps to reproduce**

1. Setting ‘ddmmyy’ => ‘d.m.y’ in LocalConfiguration
2. Let the Crawler generate indexes, if there aren't any
3. Open Info in the Sidebar and select Site-Crawler

**Environment**
- Crawler version(s): ^10, ^11
- TYPO3 version(s): 10.4.5
- Is your TYPO3 installation set up with Composer (Composer Mode): yes

**Possible Solution**

I debugged {logEntry.columns.scheduled} and this returned as my default formatted DateTime String (d.m.y). 
Maybe it is enough to remove the f:format.date(format: 'd.m.Y H:i') command. 

> ShowLog.html:
```
<f:if condition="{showResultLog}">
      <f:then>
          <td>{logEntry.columns.result_log}</td>
      </f:then>
      <f:else>
          <f:if condition="!{logEntry.columns.scheduled} || {logEntry.columns.scheduled} == '0'">
              <f:then>
                  <td>
                      <f:translate
                      key="LLL:EXT:crawler/Resources/Private/Language/locallang.xlf:labels.immediate"/>
                  </td>
              </f:then>
              <f:else>
                  <td>{logEntry.columns.scheduled-> f:format.date(format: 'd.m.Y H:i')}</td>
              </f:else>
          </f:if>
          <f:if condition="{logEntry.columns.exec_time} = '-'">
              <f:then>
                  <td>{logEntry.columns.exec_time}</td>
              </f:then>
              <f:else>
                  <td>{logEntry.columns.exec_time -> f:format.date(format: 'd.m.Y H:i')}</td>
              </f:else>
          </f:if>
      </f:else>
  </f:if>
```

[...]

PR #871, opened 2022-01-04
[BUGFIX] Fix DateTime parse error in backend module when displaying log

brotkrueml:
Resolves: #865

## Description

The patch fixes the error

    “20.12.21 14:51” could not be parsed by \DateTime constructor: DateTime::__construct()

The dates are parsed already through BackendUtility::datetime() so the variable can be displayed unchanged.

**I have**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 ### Changed
 
 ### Fixed
+* DateTime parse error in backend module when displaying log
 * PHP8 undefined array keys in CrawlerController, ConfigurationService and middlewares
 
 ### Deprecated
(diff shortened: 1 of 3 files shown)


brotkrueml:
Maybe here:

https://github.com/tomasnorre/crawler/blob/main/Classes/Backend/RequestForm/LogRequestForm.php#L369

The fallback `0` should be changed in `-` like for the `exec_time`:

https://github.com/tomasnorre/crawler/blob/main/Classes/Backend/RequestForm/LogRequestForm.php#L353

What do you think?

tomasnorre:
Thanks for looking into this @brotkrueml. I'll have to check your suggestion and test it. 
From quickly reading it, I cannot say if it's a good idea or not. 

But sounds logic to have time handle the same in both cases.

PR #876, opened 2022-01-14
[TASK] Register icons via service container

brotkrueml:
## Description

This improves the loading speed of every request in TYPO3 v11+.
In TYPO3 v10 the old way to register icons is used.

See: https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/11.4/Feature-94692-RegisteringIconsViaServiceContainer.html

**I have**


tomasnorre:
Thanks. This looks like yet another improvement.

I rebase in the hope that the Acceptance Tests problem was fix.

tomasnorre:
Unfortunately, the Acceptance tests wasn't fixed. I'll do the manual work around again..

PR #891, opened 2022-02-01
[BUGFIX] Prevent buildQueue from creating entries for non-localized pages

LukeDberg:
Prevents buildQueue from creating entries for non-existent localized pages, resulting in 404 errors. This should partially solve #584.

I don't like that so many database queries have to be made, which is certainly bad for performance, but currently I can't find any other solution. I don't think this PR is suitable for merging yet, but the idea should become clear.

**I have**

### DIFF
--- a/Classes/Service/UrlService.php
+++ b/Classes/Service/UrlService.php
@@ -20,6 +20,7 @@ namespace AOE\Crawler\Service;
  */
 
 use Psr\Http\Message\UriInterface;
+use TYPO3\CMS\Core\Database\ConnectionPool;
 use TYPO3\CMS\Core\Http\Uri;
 use TYPO3\CMS\Core\Routing\SiteMatcher;
 use TYPO3\CMS\Core\Site\Entity\Site;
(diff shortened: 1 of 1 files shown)


tomasnorre:
I haven't done the review yet, but could you please check the functional tests? 

```shell
1) AOE\Crawler\Tests\Functional\Service\UrlServiceTest::getUrlFromPageAndQueryParametersReturnExpectedUrl with data set "With PageId and QueryString (including Language (FR))" (1, 'id=21&L=1&q=crawler', '', 1, TYPO3\CMS\Core\Http\Uri Object (...))
Error: Call to a member function getHost() on null
```

LukeDberg:
> I haven't done the review yet, but could you please check the functional tests?
> 
> ```shell
> 1) AOE\Crawler\Tests\Functional\Service\UrlServiceTest::getUrlFromPageAndQueryParametersReturnExpectedUrl with data set "With PageId and QueryString (including Language (FR))" (1, 'id=21&L=1&q=crawler', '', 1, TYPO3\CMS\Core\Http\Uri Object (...))
> Error: Call to a member function getHost() on null
> ```

I don't think it's a problem with the new code, but I'll check it out.

tomasnorre:
Hi @LukeDberg 

Thanks for your contribution, I was a little fast merging. 
The 

```php 
if ($url === null) {
    continue;
}
```

Gives me troubles, for my Static Code Analyser. A string comparison against `null` will always be false. 

If I remove it, my tests are still working. 

Can you please verify if the fix works for you without the check? 
For me, it looks like it's still working. 

Please let me know.

tomasnorre:
hi @LukeDberg 
Can you please try out following instead, in the CrawlerController where the `if ($url === null)` is set right now. 

```php
$url = $urlService->getUrlFromPageAndQueryParameters(
    $pageId,
    $urlQuery,
    $vv['subCfg']['baseUrl'] ?? null,
    $vv['subCfg']['force_ssl'] ?? 0
);

if (! $url instanceof UriInterface) {
    continue;
}

$url = (string) $url;
```

LukeDberg:
Hi @tomasnorre,

tried it out. Works like a charm.

tomasnorre:
Super, thanks for testing. Then I'll merge my PR when CI is green. 
https://github.com/tomasnorre/crawler/pull/900

Issue #757, opened 2021-06-28
Crawler builds queue on 10.4 project, but doesn't process queue


Typine3:
Hello,
I'm trying to get crawler running on a 10.4 project sinces days - with no success. I'm always facing the same problem: 
I can produce (by "build" task in scheduler) records in tx_crawler_queue. 
When I want to process them (via "run" scheduler task), something happens, for example result_data and process_id is filled - but **process_id_completed stays empty and nothing is written in the index_x tables, for example index_rel.**
What I did:
- indexed_search and crawler are installed.
- In indexed_search "Debug mode" ist enabled and "Disable Indexing in Frontend" is set to true. "Use "crawler" extension to index..." can be true or not, the result is the same.
- In crawler I did not change the default configuration except filling the "Name of php binary" field (which seems to be correct - I don't get an error message in Info-SiteCrawler-CrawlingProzesse) and aktivating "debug.processDebug".
- I created a Crawler Configuration Record in root page which is used for the build task. There ist nothing special - I use "Re-indexing tx_indexedsearch_reindex" and in "Pids only" I wrote only the startpage for test.
If would be very happy if someone has an idea - I'm running out of ideas now feeling I tested everything that could be important.
Also I can exclude a special configuration in this project - as I tried everything on two different 10.4 projects hosted by different providers - the result is the same.
Thanks, CB

tomasnorre:
Hi @Typine3 

Sounds strange to me. 

Which version of the Crawler are you using? 
Are you installing you TYPO3 with composer? 
Which PHP version? 
Which MySQL/MariaDB Version?

Typine3:
crawler 9.2.5
No Composer
php7.3LATEST
MySQL 5.6.19
Thanks for answering so fast!

tomasnorre:
Thanks for the additional info. Will see if I can get to reproduce this next week.

tomasnorre:
As the Crawler works without indexed_search I tend to say it's a misconfiguration of the indexed_search. 
Could you perhaps check if there are any pointers here to help you out? 
https://t3terminal.com/blog/typo3-indexed-search/

A good way to prove if indexed_search is working as intended is to enable frontend indexing. If that works, we are a little closer to locating the problem.

Typine3:
crawler 9.2.5
No Composer
php7.3LATEST
MySQL 5.6.19
Thanks for answering so fast!

Typine3:
[...]

PR #899, opened 2022-02-10
Bugfix/trusted host pattern failing

andreashager:
## Description

The problem was that the crawler could not crawl websites in the frontend on the production server. The problem was a trusted host pattern fail due to the check of the server port that was never been set: https://github.com/TYPO3/typo3/blob/10.4/typo3/sysext/core/Classes/Utility/GeneralUtility.php#L2929

Additionally a type mismatch error on the crawler log in the backend was fixed.

Maybe this could be the same problem as described in #757 🤔 

**I have**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Changed
 
 ### Fixed
+* Fatal PHP error is thrown in the backend crawler log
 
 ### Deprecated
 #### Classes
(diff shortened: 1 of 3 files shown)


tomasnorre:
Super! Thank you for taking the time to create your first pull request. Please give us a bit of time to review it.

andreashager:
Cool thanks 🥳 
Than I can remove my local patches again 😃

Issue #515, opened 2020-01-15
sleep time and sleep after finished are swapped


allesmeins89:
`$sleeptime` and `$sleepAfterFinish` are swapped in Classes/Command/CrawlerCommandController.php in version 6.7.0 in lines 223 and 224.

`$sleeptime` is calculated with `usleep(intval($sleepTime))` that's are microseconds not milliseconds as the help says. Line 203 also in Classes/Command/CrawlerCommandController.php in version 6.7.0.

tomasnorre:
Hi @allesmeins89 

Could you please create a PR for this, then I'll be happy to test this.

tomasnorre:
This is fixed with commit https://github.com/AOEpeople/crawler/commit/618f3d8291e8d73f621d7bfaf2c4b401bedbecbc, could you please verify @allesmeins89

tomasnorre:
Fixed for typo3v9 with commit https://github.com/AOEpeople/crawler/commit/565ddcb73d739b162f033d9af58642b54f3785d9

PR #927, opened 2022-04-22
Updated descriptions of command options

schoeppe:
The descriptions for sleeptime and sleepafter where swapped (related to #515).
The unit for sleeptime is microseconds (used in usleep) not milliseconds (fixed in Command description and ext_conf_template.txt).


tomasnorre:
Thanks for your contribution @schoeppe

PR #937, opened 2022-08-08
tx_crawler_queue shows error

ste101:
When installing tx_crawler_queue is not inserted.
I see an error `'Invalid default value for 'qid''` when trying to insert tx_crawler_queue.

With this change the table will be inserted.


tomasnorre:
Thanks for your PR. Which database and version are you on? 
I'm surprised this isn't caught by our tests.

ste101:
I'm using TYPO3 10.4 and Mysql 8.0.29

tomasnorre:
Strange that the tests aren't failing for that version of MySQL in the CI. Will have to look into this.

Issue #758, opened 2021-07-13
Check for `ps` support in ProcessCleanUpHook


kpnielsen:
## Feature Request

**Is your feature request related to a problem? Please describe.**
On some shared hosting environments (e.g. Mittwald which is very popular in Germany) the `ps` command is not available due to security considerations. The [ProcessCleanUpHook](https://github.com/AOEpeople/crawler/blob/9.2.5/Classes/Hooks/ProcessCleanUpHook.php#L212) uses `ps` to find orphaned queue processes if the option `cleanup.cleanUpOldQueueEntries` is set to `true`.

**Describe the solution you would like**
Ideally the hook would check whether `ps` is  can be executed and try to execute an alternative, if it is not. I don't know of any such alternatives, though.

**Describe alternatives you've considered**
Alternatively, the hook could use `try {} catch() {}` and merely print out a warning.

tomasnorre:
Hi @kpnielsen 

Thanks for your input. I wasn't aware of that, and will try to come up with another solution on how to detect this.

kpnielsen:
Just got back to this because a colleague of mine stumbled across this very same problem, again.
It might be enough to wrap the `exec('ps aux ...')` call in the hook with an if-statement like so
```php
// …
if (`which ps`) {
    exec('…');
} else {
    trigger_error("Cannot locate ps command", E_USER_WARNING);
}
// …
```
This would at least prevent the scheduler from failing altogether. Is there anything that speaks against it?

tomasnorre:
This would at least stop the scheduler from failing, but would still not make the crawler working on systems without `ps`. 

So we can add this for now, and still consider better options. 
Would you mind creating the PR?

tomasnorre:
Sorry for the long wait @kpnielsen.

What do you think about using the `proc` files on the system instead of `ps` command? 

```php
foreach (glob('/proc/[0-9]*/status') as $file) {}
```

I think it might be two slow, as there a quite some files to process in there, and it's read from disk, not even memory. 

I'm not sure I think it's a good alternative to `ps`.

PR #945, opened 2022-10-04
Check for ps command before executing it

kpnielsen:
## Description
This PR adds a simple check to see whether the `ps` command can be found before actually executing it in the ProcessCleanUpHook. While this PR does not add any functionality to use alternative methods to find orphaned crawler processes, it does prevent the hook from throwing an exception if the `ps` command can not be found.

See: #758 .

**I have**

### DIFF
--- a/Classes/Hooks/ProcessCleanUpHook.php
+++ b/Classes/Hooks/ProcessCleanUpHook.php
@@ -192,7 +192,15 @@ class ProcessCleanUpHook implements CrawlerHookInterface
         $returnArray = [];
         if (! Environment::isWindows()) {
             // Not windows
-            exec('ps aux | grep \'typo3 crawler:processQueue\'', $returnArray, $returnValue);
+            if (`which ps`) {
+                // ps command is defined
+                exec('ps aux | grep \'typo3 crawler:processQueue\'', $returnArray, $returnValue);
+            } else {
+                trigger_error(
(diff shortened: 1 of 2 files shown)


Classes/Hooks/ProcessCleanUpHook.php
```diff
@@ -192,7 +192,15 @@ private function findDispatcherProcesses()
         $returnArray = [];
         if (! Environment::isWindows()) {
             // Not windows
-            exec('ps aux | grep \'typo3 crawler:processQueue\'', $returnArray, $returnValue);
+            if (`which ps`) {
```
tomasnorre, line 195: Shouldn't this be wrapped in an `exec()` too?

kpnielsen: This uses backticks (https://www.php.net/manual/en/language.operators.execution.php), which is equivalent to `shell_exec`. I don't mind writing it either way.

tomasnorre: I'm just thinking of keeping it the same for both commands. I would prefer the exec()

kpnielsen: You are absolutely right. I updated the PR accordingly.

tomasnorre (approved):

PR #958, opened 2022-10-11
Support environment variables as base_url

maikschneider:
To allow a crawler configuration on a per-instance basis, it would be nice to use an environment variable as alternative base_url, e.g. $TYPO3_RELEASE_URL

### DIFF
--- a/Classes/Backend/RequestForm/LogRequestForm.php
+++ b/Classes/Backend/RequestForm/LogRequestForm.php
@@ -12,10 +12,13 @@ use AOE\Crawler\Utility\MessageUtility;
 use AOE\Crawler\Value\QueueFilter;
 use AOE\Crawler\Writer\FileWriter\CsvWriter\CrawlerCsvWriter;
 use AOE\Crawler\Writer\FileWriter\CsvWriter\CsvWriterInterface;
+use Doctrine\DBAL\Driver\ResultStatement;
+use Doctrine\DBAL\ForwardCompatibility\Result;
 use Doctrine\DBAL\Query\QueryBuilder;
 use TYPO3\CMS\Backend\Tree\View\PageTreeView;
 use TYPO3\CMS\Backend\Utility\BackendUtility;
 use TYPO3\CMS\Core\Database\ConnectionPool;
+use TYPO3\CMS\Core\Exception;
 use TYPO3\CMS\Core\Imaging\Icon;
 use TYPO3\CMS\Core\Imaging\IconFactory;
 use TYPO3\CMS\Core\Utility\DebugUtility;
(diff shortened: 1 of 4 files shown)


tomasnorre:
Hi @maikschneider 

Thanks for your suggestion. I will look into this if I think it's needed or not.
 
I have always used extensions like: https://github.com/bmack/site-importer to keep environment dependent settings configured correctly.

Likewise, I will consider the implementation too if I decide to accept the suggestion. I'll get back to you.

maikschneider:
Thanks @tomasnorre for the fast feedback!

The extension site-importer is an interesting approach - I'll look into it, thanks :)

FYI: I'm used to configure my TYPO3 installations a lot via env vars, this makes it very easy during deployments and for overriding settings. There was a [discussion](https://github.com/sourcebroker/deployer-extended-typo3/discussions/7) where I explained how my `AdditionalConfiguration.php` and `.env` look like.

tomasnorre:
Thanks for the follow-up. I'll read that.

tomasnorre:
Would you mind to, try to get the CI green again. The Backwards compatibility will not be green before a 12.0.0 release, and the non-composer acceptance tests i blinky. But the rest should be green. 

Thanks in advance.

maikschneider:
Is there anything left I can do? As far as I can see, all failed checks relate to outdated test configurations..

Classes/Backend/RequestForm/LogRequestForm.php
```diff
@@ -147,8 +152,13 @@ private function showLogAction(int $setId, string $quiPath): string
...
                     )
-                    ->execute()
-                    ->fetch();
+                    ->execute();
+
+                if (!$q_entry instanceof Result) {
+                    throw new Exception('Could not find Queue entry', 1665559362);
+                }
+
+                $q_entry = $q_entry->fetch();
```
brotkrueml, line 161: I propose to avoid overwriting an existing variable with another value type as it may confuse.

Proposal: Use `$result` for the `execute()` assigment (as this should contain a result object anyway) and then call `$result->fetch()` to populate `$q_entry`.

Classes/Service/UrlService.php
```diff
@@ -81,6 +81,9 @@ public function getUrlFromPageAndQueryParameters(int $pageId, string $queryStrin
             }
             $url = $site->getRouter()->generateUri($pageId, $queryParts);
             if (!empty($alternativeBaseUrl)) {
+                if (strpos($alternativeBaseUrl, '$') === 0) {
```
brotkrueml, line 84: As crawler uses PHP8 in main, you can use `str_starts_with()` instead of `strpos()` which is better readable.

brotkrueml (commented):
A suggestion: add this feature also to the documentation, so everyone know it exists and how to use it.

Classes/Service/UrlService.php
tomasnorre: The backport isn't a problem as we are requiring the symfony/polyfill8

tomasnorre:
Just saw this is done against v11.x should be main branch. v11.x will only get bugfixes.
Could you please change your PR?

PR #1002, opened 2023-01-27
Fix: Wrapping

localheinz:
This pull request

- [x] wraps arrays in YAML files


localheinz:
Thank you, @tomasnorre!

PR #1003, opened 2023-01-27
Fix: Run `Build/bin/rector process --config=rector-ci.php`

localheinz:
This pull request

- [x] runs `Build/bin/rector process --config=rector-ci.php`


tomasnorre:
As this is a bug in the TYPO3 Rector and not in the Crawler itself, I will reject this PR.
https://github.com/sabbelasichon/typo3-rector/pull/3235

PR #1009, opened 2023-04-19
Fix _ADDTABLE behavior

Hawkeye1909:
We had issues using the _ADDTABLE parameter after upgrading an old installation from v6.7.3 to v10.0.3:

We use configurations like `[_TABLE:tx_news_domain_model_news;_ADDTABLE:, sys_category_record_mm;_PID:1;_WHERE: hidden = 0 AND tx_news_domain_model_news.uid = sys_category_record_mm.uid_foreign AND uid_local = 2]` to restrict the category for news entries found by the crawler. With your code this results in queries like `SELECT uid FROM , , WHERE (pid IN (?)) AND (hidden = 0 AND tx_news_domain_model_news.uid = sys_category_record_mm.uid_foreign AND uid_local = 1)`. Of course, this will result in an error.

With my changes it is possible to add several additional tables separated by commas. Even our old configuration would work, although the comma at the beginning is no longer needed.

This is only tested with v10.0.3 on TYPO3 v9.5.31 as the upgrade is still in progress...

## Description

**I have**


Hawkeye1909:
@tomasnorre Could you please add this fix also to the v11.x branch?

tomasnorre:
@Hawkeye1909 Yes, I can do so to. It would ease the process if you create a PR for that branch too ;)

PR #1027, opened 2024-02-13
Fix deprecation notice in BuildQueueCommand.php

olilo:
## Description

The following deprecation warning frequently occurred when using PHP8.2

PHP Deprecated:  Using ${var} in strings is deprecated, use {$var} instead

This PR fixes this deprecation notice.

See also: https://php.watch/versions/8.2/$%7Bvar%7D-string-interpolation-deprecated

**I have**


tomasnorre:
Thanks for contribution, we have all ones or twice open an issue that was a duplicate, ;) no worries.

PR #1022, opened 2023-09-01
Compatibility for TYPO3 12.4

IndyIndyIndy:
## Description

Gets everything running for TYPO3 12.4.
Unit Tests are working now, but I couldn't get the functional/acceptance tests running and am out of time.

**I have**

- [/ ] Checked that CGL are followed
- [ ] Checked that the Tests are still working
- [ ] Added description to CHANGELOG.md (github-handle is optional)
- [ ] Added tests for the new code

PR #1058, opened 2024-04-30
fix 'invalid omit_containers'

DavidBruchmann:
## Description

this commit tries to fix failing of `Run jonaseberle/github-action-setup-ddev@v1` in #1022, by defining containers to omit differently.

if the mentioned run above is working now the PR could be merged, else just dump it, or remove `db` in `omit_containers` from `.devbox/.ddev/config.yaml`.

Note, that the whole failing tests include still another issue, which is not addressed in this commit / PR.

### DIFF
--- a/.devbox/.ddev/config.yaml
+++ b/.devbox/.ddev/config.yaml
@@ -26,7 +26,7 @@ hooks:
     - exec: /var/www/html/bin/typo3 cache:flush
     - exec: /var/www/html/bin/typo3 cache:warmup
     - exec: /var/www/html/bin/typo3 backend:unlock
-omit_containers: [ dba, ddev-ssh-agent ]
+omit_containers: [ db, ddev-ssh-agent ]
 webimage_extra_packages: [ cron ]
 use_dns_when_possible: true
 timezone: Europe/Copenhagen
--- a/.devbox/.ddev/ddev.site/config.nodb.yaml
+++ b/.devbox/.ddev/ddev.site/config.nodb.yaml
@@ -1,0 +1,1 @@
+omit_containers: [ db, ddev-ssh-agent ]


DavidBruchmann:
The linked issue and the docs shine a light on this issue:

 - https://github.com/ddev/ddev/issues/3718
 - https://ddev.readthedocs.io/en/stable/users/configuration/config/#omit_containers
 - https://ddev.readthedocs.io/en/stable/users/configuration/config/#project_tld

DavidBruchmann:
Great that it worked, thanks for the feedback!

tomasnorre:
It's merged, but failed, due to a problem with ddev-github-action atm. 
https://github.com/ddev/github-action-setup-ddev/issues/27

DavidBruchmann:
Yes, I know ;-)

> Note, that the whole failing tests include still another issue, which is not addressed in this commit / PR.

DavidBruchmann:
I'm not sure if adding db there is correct, see here: https://github.com/tomasnorre/crawler/actions/runs/8892439954/job/24416494000#step:7:1105

tomasnorre:
I was correct. As the

> Yes, I know ;-)
> 
> > Note, that the whole failing tests include still another issue, which is not addressed in this commit / PR.

I overlooked that comment.

>  I'm not sure if adding db there is correct, see here: https://github.com/tomasnorre/crawler/actions/runs/8892439954/job/24416494000#step:7:1105

The DB setup should be right, though, but will test when I find the time and energy.

tomasnorre:
There was a regression in the fix I merged of yours. 
https://github.com/tomasnorre/crawler/commit/8d72fead4c8fac0150994b7162e3a901343496e4

The reason it worked before with the wrong "dba" in omit_containers, was that it was never meant to be omitted. Therefore it worked before :)

I have adjusted the ddev configs, to avoid confusions like this in the future. Thanks for your contribution, without it, I would have found the problem any time soon.

DavidBruchmann:
Great that you tackled it!
Regrettable only a small contribution from my side ;-)

tomasnorre:
Every contribution has value.

Issue #1082, opened 2024-06-26
Process Queue throws Truncated incorrect DECIMAL value


xyng:
## Bug Report

**Current Behavior**
Running the command `crawler:processQueue` ends with exception: `Truncated incorrect DECIMAL value: 'VALUE_HERE'`

**Expected behavior/output**
Running the command does not throw exception.

**Steps to reproduce**
Run command `crawler:processQueue`.

**Environment**
- Crawler version(s): master, 12.0.3
- TYPO3 version(s): 12.4.16
- PHP version(s): 8.3
- Is your TYPO3 installation set up with Composer (Composer Mode): yes

**Possible Solution**
https://github.com/tomasnorre/crawler/blob/68e72580ad300ab8ac545daf64c62f62cc897871/Classes/Domain/Repository/QueueRepository.php#L330
Use `ArrayParameterType::STRING`.
The field `process_id` is defined as `varchar`:
https://github.com/tomasnorre/crawler/blob/1606131d50a7bc6e0676f8a30a92f6b52f5de605/ext_tables.sql#L15
https://github.com/tomasnorre/crawler/blob/68e72580ad300ab8ac545daf64c62f62cc897871/ext_tables.sql#L34

I can send a PR, but since this is probably a one-line change, it may be more time-consuming for you.
Please tell me if you want me to make one.

**Additional context**
-

tomasnorre:
Thank you for your report. Do you get this exception every time running the command? 
I have never paid attention to it.

If you make a PR I would be happy to merge it after evaluating it. I cannot say from the top of my head if the solution is the correct one, would need to dig a bit, but sounds correct from reading it. But would need to look at possible side effects.

Best case would be if you can add a test that proves that the exception happens and disappears after the fix.

xyng:
Hi @tomasnorre - thanks for your reply.

I started to see this error every few minutes from our scheduler and decided to investigate.
I then found this type mismatch. Since the `process_id` is stored as a varchar and the contents seem to be md5-values, and integer will fail to be converted.

I will prepare a PR and also try to create a test.

tomasnorre:
Resolved by https://github.com/tomasnorre/crawler/pull/1083

PR #1083, opened 2024-07-08
Fix: ProcessQueue throws truncated incorrect decimal value

xyng:
## Description

This PR provides a test and fix for #1082 - changes the param-type of `processId` in the prepared statement to `string` instead of `integer`, since the database field already uses `varchar(50)` and `ProcessQueueCommand` will pass a string.
That can lead to either non-matches (sqlite) or even runtime errors (in one of our environments using MariaDB).

**I have**

## Note

@tomasnorre I tried my best to get the test to reproduce. For some reason, only the SQLite one ran on my machine. For SQLite, there is no runtime error, but non-numeric values are not matched, leading the test case to fail the count-check which is somewhat sufficient to evaluate the problem.
I'd appreciate if you could check the MySQL and PostgreSQL derivates if they work correctly (either throw and exception or at least fail the count test).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Changed
 
 ### Fixed
+* Prevent exception `Truncated incorrect DECIMAL value` in `crawler:processQueue` [@xyng](https://github.com/xyng)
 
 ### Deprecated
 
--- a/Classes/Domain/Repository/QueueRepository.php
+++ b/Classes/Domain/Repository/QueueRepository.php
@@ -327,7 +327,7 @@ class QueueRepository extends Repository implements LoggerAwareInterface
                 $queryBuilder->expr()->eq('exec_time', $queryBuilder->createNamedParameter(0, Connection::PARAM_INT)),
                 $queryBuilder->expr()->in(
                     'process_id',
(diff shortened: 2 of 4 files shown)


tomasnorre:
@xyng Thank you so much for your contribution. Looks good at first glance. Will have the CI check the MySQL, MariaDB and PostgreSQL for us ;) 

There are some code-style, use statement sorting not right atm, but I think that is due to a rector update that isn't caused by you. So don't worry about those if the pipeline turns red.

xyng:
@tomasnorre thank you for checking and your feedback!
I have added an entry to the changelog - not entirely sure about the wording.

tomasnorre:
> @tomasnorre thank you for checking and your feedback! I have added an entry to the changelog - not entirely sure about the wording.

The wording is fine. :)

tomasnorre:
@xyng Thanks for your contribution, cannot say when it will be released.

graphicfox:
Hello, is there any update when this will be released? My scheduler task never runs because of this error. :(

tomasnorre:
Hi.. 

I can make a release some when next week. Just need a little preparation.

PR #1088, opened 2024-07-23
fix "crawler:buildQueue fails with Error at offset 0 of 92 bytes"

### LINKED ISSUE #1087: crawler:buildQueue fails with Error at offset 0 of 92 bytes
## Bug Report

**Current Behavior**

i execute the following command to build a queue and crawl the found pages.
"typo3cms crawler:buildQueue --depth 3 --mode exec 1 default"
After list of the found urls i get 

>Processing
>
>   0/299 [>---------------------------]   0%
>Error at offset 0 of 92 bytes

**Expected behavior/output**
Execution should not fail

**Steps to reproduce**

Prepare a configuration
execute "typo3cms crawler:buildQueue --depth 3 --mode exec 1 default"

**Environment**

- Crawler version(s): 11.0.10
- TYPO3 version(s): 11.5.38
- PHP version(s): 8.3

**Possible Solution**
The error appiers in JsonCompatibilityConverter line 39.

The dataString for unserialize in my behavior is 
{"url":"https:\/\/www.domain.de\/en\/home.html","procInstructions":[""],"procInstrParams":[]}

This is not valid for unserialize. When i comment the lines 39-49 the execution works.

stephan-brun:
Fix #1087

### DIFF
--- a/Classes/Converter/JsonCompatibilityConverter.php
+++ b/Classes/Converter/JsonCompatibilityConverter.php
@@ -36,6 +36,11 @@ class JsonCompatibilityConverter
      */
     public function convert(string $dataString)
     {
+        $decoded = json_decode($dataString, true);
+        if ($decoded) {
+            return $decoded;
+        }
+
         $unserialized = unserialize($dataString, ['allowed_classes' => false]);
         if (is_object($unserialized)) {
             throw new \Exception('Objects are not allowed: ' . var_export($unserialized, true), 1593758307);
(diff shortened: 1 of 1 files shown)


tomasnorre:
First, thanks a lot for providing the PR.
To be honest. I have see this behavior before and tried to change the order of the if-checks, but gave other side effects. I don't recall from the top of my head, but some unwanted effects was present. I need to dig into this a little more again to validate the PR.

From a logical perspective, your check first sounds reasonable, as the converter is only there, is old data format is still present in queue, which shouldn't happen that often, only directly after an update. 

So perhaps in Idea that I should consider implement, in the dev-main branch, as v11.x only get bugs fixes. Would be to make a command that updates the format ones, and remove the converter.

What do you think @stephan-brun @hacksch ?

cweiske:
@tomasnorre you can close this PR; the problem has been solved with #1117.
But please tag the latest v11.x branch commit.

tomasnorre:
Resolved in #1117

PR #1116, opened 2024-12-03
[BUGFIX] Use proper pid's in EXT:news docs code example

Patta:
The pid's in the EXT:news docs code example are incorrect. This PR sets the pid's in the code example as shown in the screenshot.


tomasnorre:
Thanks for your contribution.

PR #1117, opened 2025-01-14
[BUGFIX] Catch E_WARNING when unserialize() fails

cweiske:
Since PHP 8.3, unserialize() throws an E_WARNING instead of E_NOTICE. This leads to an error when viewing crawler information for pages and when crawling sites:

> PHP Warning: unserialize(): Error at offset 0 of 122 bytes
> in typo3conf/ext/crawler/Classes/Converter/JsonCompatibilityConverter.php line 39

This patch solves the problem in the most non-invasive way by simply catching the unserialize warning.

Resolves: https://github.com/tomasnorre/crawler/issues/1087

## Description

**I have**

### DIFF
--- a/Classes/Converter/JsonCompatibilityConverter.php
+++ b/Classes/Converter/JsonCompatibilityConverter.php
@@ -36,7 +36,11 @@ class JsonCompatibilityConverter
      */
     public function convert(string $dataString)
     {
-        $unserialized = unserialize($dataString, ['allowed_classes' => false]);
+        try {
+            $unserialized = unserialize($dataString, ['allowed_classes' => false]);
+        } catch (\Throwable $e) {
+            $unserialized = false;
+        }
         if (is_object($unserialized)) {
             throw new \Exception('Objects are not allowed: ' . var_export($unserialized, true), 1593758307);
         }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Fixed
 * Fixed query building when using _ADDTABLE parameter [@hawkeye1909](https://github.com/hawkeye1909)
 * Typecast PageTS Crawler Configuration `force_ssl` to `int`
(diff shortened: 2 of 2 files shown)


tomasnorre:
Thanks for your PR. I'll review one of the next days.

Classes/Converter/JsonCompatibilityConverter.php
```diff
@@ -36,7 +36,11 @@ class JsonCompatibilityConverter
      */
     public function convert(string $dataString)
     {
-        $unserialized = unserialize($dataString, ['allowed_classes' => false]);
+        try {
+            $unserialized = unserialize($dataString, ['allowed_classes' => false]);
+        } catch (\Throwable) {
```
tomasnorre, line 41: I know you're not doing anything with the exception, which is find, but it needs to have a variable too, for it to being valid PHP code. That's also why a lot of the pipeline fails. 

```php
} catch (\Throwable $exception) {
```

Can you please correct that, thanks.

cweiske: - The actions i've checked all fail because of "Error: This request has been automatically failed because it uses a deprecated version of `actions/upload-artifact: v2`" and not php code
- Since PHP 8 you don't need a variable anymore in the catch block: https://wiki.php.net/rfc/non-capturing_catches - but since v11.x still supports php 7.4 I'll add it.

tomasnorre: Thanks. I didn't think of the diff from 7.4 to 8.0 I installed PHP 7.4 to check it out, and therefor saw the problem. 

The "tests" I meant the Static Analysis ones.

Thanks for taking care.

cweiske:
Version [11.0.11](https://github.com/tomasnorre/crawler/releases/tag/11.0.11) has the fix.

PR #1121, opened 2025-01-29
[BUGFIX] Fix showing "Scheduled" and "Run-time" columns in backend log

cweiske:
In Info > Site Crawler > Crawler log:
As soon as the backend user activated the "Show Result Log" checkbox one time, the columns "Scheduled" and "Run-time" are empty when the checkbox is disabled.

The reason is that the variable's value is not taken into account, only its existence in the user session data.

## Description

**I have**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog TYPO3 Crawler
 
+## Crawler 11.0.12-dev
+
+### Fixed
+* "Scheduled" and "Run-time" column values in backend log [cweiske](https://github.com/cweiske)
+
 ## Crawler 11.0.11
 
 Crawler 11.0.11 was released on January 20th, 2025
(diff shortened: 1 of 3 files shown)


tomasnorre:
Thanks for the PR. 

Would you mind updating the PHPStan baseline, so that it runs green again. 
I have just re-triggered the Acceptance Tests, as there should be no reason for them to fail. 

Would this also apply for 12.0.x releases? 

I sort of had decided to not do anymore bug fixes for the 11.x branch as TYPO3 11 it ELTS now. 
But if it requires less work from me, I'll accept it.

cweiske:
- rebased
- phpstan fixed

> Would this also apply for 12.0.x releases?

No, the backend log view had been rewritten in 2022 with commit 0e7d7e29133c33a598aa1b4c98fbf6cecf723af8.

PR #1132, opened 2025-02-11
[FEATURE] Add getPageFromQueue to QueueService

tomasnorre:
This is WIP to see if I can find a better way to generate the QueueRows needed in e.g. the BuildQueueCommand. 

If this works out well, quite some code from the CrawlerController can be deprecated. Fingers crossed.


tomasnorre:
This will not get merge this time. Will look at it later again.

PR #1142, opened 2025-05-12
[BUGFIX] Fix indexing of cached pages when "disableFrontendIndexing" …

cweiske:
…is active

When pages are cached in TYPO3v12, the indexing process is not triggered when crawler fetches the pages.
This becomes a problem when the "Disable Indexing in Frontend" (basic.disableFrontendIndexing) indexed_search flag is enabled, because no indexing is happening then.

The solution is to tell TYPO3 to not use the cache when the page is requested by the crawler process.

See the ShouldUseCachedPageDataIfAvailableEvent documentation at https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/12.0/Feature-96968-PSR-14EventForAvoidLoadingFrontendPagesFromCache.html

Resolves: https://github.com/tomasnorre/crawler/issues/1029

**I have**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * Adjust path in cli/bootstrap.php to match new folder structure of where extensions are installed
 * Don't wrap empty HTML in table-tag to ensure valid HTML
 * Fix crawling of mount pages
+* Fix indexing of cached pages when "disableFrontendIndexing" is active [@cweiske](https://github.com/cweiske)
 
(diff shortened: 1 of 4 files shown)


tomasnorre:
Thank you for your PR @cweiske, could you please try to fix the PHPStan issue in CI? 
The other tests are little more blinkers, so not always green. 

Thanks.

cweiske:
I adjusted the Process.php file

Issue #1137, opened 2025-03-07
Creation of dynamic property TYPO3\CMS\Frontend\Controller\TypoScriptFrontendController::$applicationData is deprecated


chludwig:
## Bug Report

**Current Behavior**
In the system log there are hundreds off error messages each day:
> Core: Exception handler (WEB): Uncaught TYPO3 Exception: #1476107295: PHP Runtime Deprecation Notice: Creation of dynamic property TYPO3\CMS\Frontend\Controller\TypoScriptFrontendController::$applicationData is deprecated in /home/www/production/vendor/tomasnorre/crawler/Classes/Middleware/CrawlerInitialization.php line 60 | TYPO3\CMS\Core\Error\Exception thrown in file /home/www/roduction/vendor/typo3/cms-core/Classes/Error/ErrorHandler.php in line 141. Requested URL: https://domain.de/dbsystel

They are produced by the scheduled crawler.  
![Image](https://github.com/user-attachments/assets/2eebfd28-9b27-4abf-85f9-5f8ad0e99771)

**Expected behavior/output**
Crawling shouldn't produce errors.

**Steps to reproduce**
Configure a crawler as shown above and execute it. (The website was updated from TYPO3 v12.4)

**Environment**

- Crawler version(s): 12.0.7
- TYPO3 version(s): 13.4.5
- PHP version(s): 8.3.16
- Is your TYPO3 installation set up with Composer (Composer Mode): yes

maelstromer:
While upgrade a client TYPO3 system to version 13.4 I have encountered the same issue. The TYPO3 log produces the same TYPO3 Exception that chludwig postig in the OP, for each page in the processing-queue.

I had a look at the CrawlerInitialization class and noticed that it attempts to access the property `$GLOBALS['TSFE']->applicationData` multiple times, starting at line 60. However, according to "Breaking #102600" in the TYPO3 changelog, this property was removed with TYPO3 v13 (see: https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/13.0/Breaking-102600-TSFE-applicationDataRemoved.html#breaking-102600-tsfe-applicationdata-removed)

This would explain the error message, because it is being interpreted as an attempt to dynamically create the property, due to it no longer being part of `$GLOBALS['TSFE']`.

When processing the built queue via CLI, no new entries are added to the index and crawler responds with the following:
`Unprocessed Items remaining:16 (aa786c059902b47a15869dd96f48767e)
3`

[...]

PR #1158, opened 2025-08-20
!!! [TASK] Replace deprecated applicationData with request attributes

tomasnorre:
## Description

This is a PR idea that might solve the problem with issue #1137 at the end, this is still rough WIP, but wanted some feedback.

### Breaking Change
This will result in a breaking change, as the runPollSuccessHooks hook changes. See documentation changes for more info.

**I have**


tomasnorre:
@cweiske As you are currently quite active, which I highly appreciate, do you have an suggestion/feedback/opinion on this PR? 

And if you like the approach could you help me get the tests green again?

cweiske:
At first I thought that a simple variable in the CrawlerInitialization class could be used instead of applicationData, but as I see it, the problem is that other extensions have to be able to communicate their success to crawler somehow.
Writing applicationData was an easy way to do that without having to know if crawler is installed at all.

With the removal of `applicationData`, extensions must to be adjusted, in one way or another. In this case, crawler could provide an API method for the other extensions and would not need to rely on the request attributes at all.

My suggestion would be a static method e.g. `CrawlerInitialization::extensionSuccess($key, $status)` that other extensions would need to call instead setting `$GLOBALS['TSFE']->applicationData['tx_crawler']['success'][$key] = true`. This would need to be wrapped in a `if (class_exists())` call of course.

tomasnorre:
Thanks for your input, I'll take that into consideration. I don't have a solution/final idea myself yet.

PR #1159, opened 2025-08-21
[BUGFIX] Set guzzle connect timeout to 5 seconds

cweiske:
## Description
Set guzzle connect timeout to 5 seconds.

Default connect timeout behavior is to wait indefinitely, which means that crawler would hang forever in case the server has problems.

https://docs.guzzlephp.org/en/stable/request-options.html#connect-timeout

**I have**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Changed
 
 ### Fixed
+* Set guzzle connect timeout to 5 seconds [@cweiske](https://github.com/cweiske)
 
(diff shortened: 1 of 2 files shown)


tomasnorre:
Thanks for the PR. 

What do you think about this being configurable in the Extension Settings? Or is 5 seconds a sane timeout?

tomasnorre:
I see, then I think it's fine not to have it configured in the extension. I can change that later if it gives any problems.

PR #1162, opened 2025-08-27
[FEATURE] Extract error information in SubProcess crawl strategy (TYPO3v11)

cweiske:
Backport of #1161 to v11.

----
Up to now, the sub process execution strategy did not detect errors. The HTML of error pages was provided as content, and the backend showed an "OK" status.

Since cli/bootstrap.php does not set a non-zero exit code on errors, we have to parse the HTML ourselves to detect if an error occured.

This patch supports TYPO3 error pages and extracts HTTP status code, status message and error message from it.

Error information is logged into the debug log as well as made visible in the backend site crawler log view.

## Description

**I have**


cweiske:
Ah I see that v11 will get security fixes only. Then this has no chance and can be closed.

tomasnorre:
Can you please rebase, I have updated the actions/cache version to working version again.

cweiske:
phpstan is fixed now.

PR #1165, opened 2025-08-29
[BUGFIX] Clear page cache when crawling for re-indexing

cweiske:
## Description
When "basic.disableFrontendIndexing" is activated, TYPO3 does not index crawled pages when they are already cached.

This was fixed for TYPO3v12+ in [1] by using the "ShouldUseCachedPageDataIfAvailableEvent".
That event is not available in TYPO3v11.

This patch simply clears the frontend cache of a page directly before crawling it.

[1]: https://github.com/tomasnorre/crawler/pull/1142
Related: https://github.com/tomasnorre/crawler/issues/1029

**I have**


tomasnorre:
@cweiske  Do you have an idea why the jobs aren't picked up by a GHA Runner? I have tried to restart them, but with no luck.

cweiske:
The "Acceptance Tests" and "DBMS Compat Checks" here all show
> Cancelled after 1440m

When clicking on the details, we see
> The job has exceeded the maximum execution time while awaiting a runner for 24h0m0s

Someone else [had this problem and fixed it](https://github.com/google/material-design-icons/issues/1902#issuecomment-3119809355) by switching from `ubuntu-20.04` to `ubuntu-latest`. Not sure if this is relevant here.

cweiske:
Updating from ubuntu 20.04 to 24.04 helped. The runners work again.
The scrutinizer error is fixed by removing scrutinizer just like it was done in main.

The only problem left is the "Roave Backward Compatibility Check" which pulls in `roave/security-advisories`, which does not like typo3 v10 anymore.

PR #1167, opened 2025-09-02
[BUGFIX] Prevent warning when guzzle crawl requests return errors

cweiske:
## Description
Let the guzzle crawl strategy return errors in the same way that the subprocess strategy does: As array inside an 'errorlog' array.

Also rework the logic that JsonCompatibilityConverter uses to decide when to `unserialize()`.
And fix some phpunit deprecations, and document the crawl strategy return value structure.

Resolves: https://github.com/tomasnorre/crawler/issues/1166

**I have**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ### Fixed
 * Set guzzle connect timeout to 5 seconds [@cweiske](https://github.com/cweiske)
+* Prevent warning when guzzle crawl requests return errors [@cweiske](https://github.com/cweiske)
 
 ### Deprecated
 #### Classes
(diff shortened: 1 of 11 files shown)


cweiske:
@tomasnorre only one action fails now, coveralls:
> Submitting to https://coveralls.io/api/v1/jobs
> Client error occurred. status: 422 Unprocessable Entity
> Couldn't find a repository matching this job.

tomasnorre:
Thank you so much @cweiske, the Test Coveralls, blinks from time to time, but you added extra test case to the new code, so I'm fine with merging as all the tests, functional,unit and acceptance are all green.

cweiske:
@tomasnorre I don't like that you squashed all 4 commits into a single one, even dropping the commit messages. All the commits were sensible standalone ones.

tomasnorre:
@cweiske Sorry. I didn't read them like that. I'll pay attention to that in the future. 
I more or less always squash commits. 

Can I do anything to correct the mistake?

cweiske:
You could revert the last commit or reset it to the previous state and merge the 4 commits from #1169.

PR #1171, opened 2025-09-05
[FEATURE] Expose error when crawl response has no X-T3Crawler-Meta header

cweiske:
## Description
When the web server filters incoming and outgoing HTTP headers, the `X-T3Crawler-Meta` header (needed to transport crawler information back from indexing) may also get filtered away.

This patch lets crawler detect that problem and show it in the backend process log.

**I have**


cweiske:
Documentation is also updated now.

tomasnorre:
Thank you so much for all your contributions.

Is squashed commit ok for this?

PR #1175, opened 2025-09-08
[TASK] Refactor JsonCompatibilityConverter, change return type to always arrays

tomasnorre:
## Description

This is a refactoring only have one return type for the JsonCompatibilityConverter, so that it's always return arrays, but an empty array on errors, that earlier was false.

**I have**


tomasnorre:
I think you might be right. I'll make an issue to remove it when I start Crawler version 13. It would be breaking change in version 12, which I don't want. I don't want 13 before dropping support for 12 in March if I can avoid it. 

Thanks for your feedback.

PR #1309, opened 2026-08-18
Fix TYPO3 report "Extension key missing"

jonaseberle:
## Description

Sorry, I earlier pointed the PR to my repo so no wonder you have not seen it ;)

This fixes a very misleading warning in TYPO3 reports.
image
(screenshot is from fal_protect where the same situation was present)

In this case it means that this check is failing

https://github.com/TYPO3/typo3/blob/14.3/typo3/sysext/extensionmanager/Classes/Package/ComposerDeficitDetector.php#L86

The composer description, as seemingly expected now by TYPO3, should be `<ext_emconf.php title> - <ext_emconf.php description>`. I haven't found that in the docs, though.

I think the check is very questionable, but in order to silence the warning we just have to adjust the composer description.

**I have**


tomasnorre:
Thanks for letting me know. Haven't paid attention to that.

